PR #23, opened 2019-02-12
Truncate VIN and add display_name for name()

### LINKED ISSUE #22: Truncate VIN number for uniq_name
Given many HA configs are publicly hosted, posting the full VIN is probably not a good privacy setting.

alandtse:
This is a breaking change on HA as it will change the HA entity name. I want to keep this open for any comments prior to rolling this out
Resolves #22

### DIFF
--- a/teslajsonpy/vehicle.py
+++ b/teslajsonpy/vehicle.py
@@ -17,7 +17,7 @@ class VehicleDevice:
 
     def _uniq_name(self):
         return 'Tesla Model {} {} {}'.format(
-            str(self._vin[3]).upper(), self._vin[-6:], self.type)
+            str(self._vin[3]).upper(), self._vin[-3:], self.type)
 
     def id(self):
         return self._id
--- a/teslajsonpy/vehicle.py
+++ b/teslajsonpy/vehicle.py
@@ -2,18 +2,22 @@ class VehicleDevice:
     def __init__(self, data, controller):
         self._id = data['id']
         self._vehicle_id = data['vehicle_id']
+        self._display_name = data['display_name']
         self._vin = data['vin']
         self._state = data['state']
         self._controller = controller
         self.should_poll = True
 
     def _name(self):
-        return 'Tesla Model {} {}'.format(
-            str(self._vin[3]).upper(), self.type)
+        return ('{} {}'.format(self._display_name, self.type) if
+                self._display_name is not None and
+                self._display_name != self._vin[-6:]
+                else 'Tesla Model {} {}'.format(str(self._vin[3]).upper(),
+                                                self.type))
 
     def _uniq_name(self):
         return 'Tesla Model {} {} {}'.format(
-            str(self._vin[3]).upper(), self._vin, self.type)
+            str(self._vin[3]).upper(), self._vin[-6:], self.type)
 
     def id(self):
         return self._id


OverloadUT:
Thank you for taking on this change. I will test it locally in a dev HA.

alandtse:
Since this will break the entity_ids, I'm also tempted to remove the redundant type from the end of all the entities.  e.g., `binary_sensor.tesla_model_3_XXXXXX_charger_sensor`. As HA will list the domain anyway, it's always bothered me. 

I'll plan to also have it drop that last type info.
`binary_sensor.tesla_model_3_XXXXXX_charger`

Comments are welcome.

teslajsonpy/vehicle.py
```diff
@@ -2,18 +2,22 @@ class VehicleDevice:
...
         self.should_poll = True
 
     def _name(self):
-        return 'Tesla Model {} {}'.format(
-            str(self._vin[3]).upper(), self.type)
+        return ('{} {}'.format(self._display_name, self.type) if
+                self._display_name is not None and
+                self._display_name != self._vin[-6:]
+                else 'Tesla Model {} {}'.format(str(self._vin[3]).upper(),
+                                                self.type))
```
zabuldon: Maybe it is better to calculate some hash of it? As it is really bad idea to disclose this item.

alandtse: I think it's useful for a user to be able to match the ID to some value they know since some will have multiple vehicles. I think we could truncate it further so it's not the full serial number.  I doubt people will see a collision at 3-4 digits.  
Here's a vin decoding link for [reference](https://www.autocheck.com/vehiclehistory/autocheck/en/vinbasics).

OverloadUT:
I think the approach here should be different. This plugin was written a long time ago, before Home Assistant totally switched the way entity IDs are supposed to work.

Here's how it should be, based on latest HA best practices:
1. This plugin should return some sort of unique ID that the HA component will use _behind the scenes_ - this will never be in a public config file (it's in the entity registry). I recommend the last 5 digits of the VIN appended with the name of the entity, much like we already do. On the HA side, this will be used to set the `unique_id` of every entity.
2. This plugin should return a _friendly name_ of every sensor, which should use the user's name of the car. So for my car named Marcy, it should show "Marcy locks" "Marcy charging" or similar. If there is no name set for the car, default back to "Model 3" or "Model S." On the HA side, this will be used to set the `name` of the entity, which HA will use to _automatically generate the entity_id`.

This is the new way of doing things: the component only sets a friendly name which HA uses to create the entity_id. So long as unique_id is present, the user can freely change the entity_id right in the UI while HA is running.

The end result here is we don't have to worry about any part of the VIN being exposed in published configs, entity_ids are more intuitive to work with (as they won't have any part of the VIN or a hash in them), and we get all of the added benefits of the features that come with unique_id being set.

(At a later date we'll talk about converting the Tesla component to use config entries, but that's a whole different ballgame ;) )

OverloadUT:
Followup: I see now that it appears that you might already be headed down this same path that I outlined (without seeing the HA changes it's hard to be sure!)

teslajsonpy/vehicle.py
OverloadUT: Because this is the name that will display in Home Assistant (and therefore generate the entity_id), I think you should remove the VIN from this entirely. The end result should simply be "Tesla Model X Door Locks" or similar, if there is no `_display_name`. The VIN will be used for the `unique_id` and the user will be able to change the display name to whatever they want.

This suggestion would solve the problem that zabuldon highlights above of not exposing the VIN.

alandtse: @OverloadUT While it may not be a sizable population, there are users with multiple vehicles where that would result in a name collision, particularly as the Model 3 becomes more affordable. I'm inclined to just do the last 3 chars of the VIN for the unique ID and for the name if none have been set.

OverloadUT:
Regarding using the VIN in the name to avoid conflicts: One point I'm trying to make is that is no longer necessary in Home Assistant. The `unique_id` will handle the component knowing the difference between them. And in the rare case that someone has two of the same model AND they don't have a custom name set for either car, Home Assistant will simply append `_2` to the second car's entities, which the user can easily change in the UI if they choose.

All that being said, I just acknowledged that this would only happen if the user hasn't set a custom name for both vehicles, which will likely almost never happen, so it's not really a point worth losing sleep over.

teslajsonpy/vehicle.py
zabuldon: @alandtse @OverloadUT Guys, what do you think if we still use naming on "Library" side, but allow users to configure names of Tesla's on HA side? I saw this behavior with insteon_local components. They allow you to rename thing when you set it up.

zabuldon:
@OverloadUT @alandtse In my opinion now we have to decide and select one of the next:
1. Don't show ANY part of VIN
or
2. Don't change naming
or
3. Replace VIN with hash (optionally?)

I really don't think it is necessary to handle VIN as part of the name, but in some cases it can be useful. In my opinion it should be something like (hash(VIN)[:16]) at least for debug purposes.

alandtse:
I vote truncated VIN for uniq ID, nothing in name. That's what's in the code.

teslajsonpy/vehicle.py
OverloadUT: @zabuldon Exposing a unique_id to Home Assistant gives the user all of the power. They can change the display names and entity_ids of everything that has a unique_id at runtime without needing to even restart. So we get that for free :)

alandtse:
Closing for now so I can reset the staging branch. Will reopen this for v.0.25.

Issue #47, opened 2019-10-26
Authentication method causes logouts in apps


wernerb:
I'm using the library using home-assistant integration and works fine. However, after using this integration I notice i'm being forcefully logged out of my android app every once in a while which is annoying when I want to unlock the car :) 

This makes me believe the repeated authentications is rate-limited in some way, or at least, there are protections in place.

Instead of re-authenticating with the user/pass every time, we should use the refresh_token instead to fix this behaviour. 

https://tesla-api.timdorr.com/api-basics/authentication#post-oauth-token-grant_type-refresh_token

alandtse:
Not sure it's the cause of the relogins (I've always thought Tesla forces a relogin every month or so), but I have a oauth version in the works which will happen at the same time as async. Need to clean it up and submit the PR here and in HA.

Issue #24, opened 2019-02-13
Generalize API and remove home-assistant specific architecture


alandtse:
Right now, this library creates entities specifically for HA to consume.  This has two downsides that I can see.

1. Many functions are not intuitive as they assume some knowledge of the HA architecture. It took me a while to figure out that `controller.list_vehicles()` returned a list of objects for HA and not just a list of vehicles. It also means people not using HA likely won't use this library which runs against the philosophy from HA requiring the library be put in pypi to allow others to reuse.  
2. Feature updates to HA for Tesla require changes to both teslajsonpy and HA.  For example, if someone wants to make an additional sensor for charging rate, they will need to modify the sensor in this library and then make an equivalent edit to process the new data.  This impacted my ability to make changes in HA and probably is impacting others who want to tweak it.  If this library is just intended to allow access to the Tesla API, then teslajsonpy changes will only be needed if new endpoints/commands are created.

I think we should move the HA related architecture into the HA component. I'd be happy to take that project on.  

From an initial review, controller, connection, and exception make sense to me in this library.  Everything based on vehicle probably should go into HA.  

@zabuldon, this is your project to start, do you have any objections to this rearchitecture? I'll plan to start on doing that unless you object.

zabuldon:
@alandtse i'm not sure about moving into HA as an initial step.
In my opinion is much better to just generalize API and keep backward capability while we are working for a refactoring/extending.

I would like to suggest next plan:
1. Make "master" stable with current interfaces and merge only functionality not affecting cpabilities
2. Re-work architecture
3. Add tests/CI
4. Release 2-3 version with  backward capability layer (while we are still working for HA adoption)
5. Replace master with re-architected version as soon as we get ready with all changes in HA.

What do you think?

alandtse:
That seems fine. I would suggest that breaking changes (e.g., #23) all be saved for step 5. 

I'll get started on it.

OverloadUT:
Just tossing in a note here that I wholeheartedly agree with this plan. When I went to work on this component a few months back I ran in to the exact same confusion, where there was a bunch of HA-specific logic sitting in this library.

zabuldon:
One problem we should keep in mind. Customer may have a few Tesla's with same model. It was the reason why I've added vin as a part of the name.

alandtse:
[...]

PR #48, opened 2019-10-28
feat: convert to async

### LINKED ISSUE #44: Swap to async
I can probably take a look at it since I had to do it for [alexa_media_player](https://github.com/custom-components/alexa_media_player).

alandtse:
This version contains numerous changes in addition to the async conversion. It includes a new charging sensor, support for oauth, performance tweaks, and incorporation of black for formatting.
BREAKING CHANGE: API calls now require async
Closes #44, #47

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -3,7 +3,7 @@ url = "https://pypi.python.org/simple"
 verify_ssl = true
 
 [dev-packages]
-"flake8" = "*"
+flake8 = "*"
 detox = "*"
 mypy = "*"
 pydocstyle = "*"
@@ -12,6 +12,10 @@ pytest-cov = "*"
 tox = "*"
 twine = "*"
 python-semantic-release = "*"
+black = "*"
 
 [packages]
-requests = "*"
+aiohttp = "*"
+
+[pipenv]
+allow_prereleases = true
--- a/pylintrc
+++ b/pylintrc
@@ -1,10 +1,15 @@
+[MASTER]
+ignore=tests
+
 [MESSAGES CONTROL]
 # Reasons disabled:
+# format - handled by black
 # unnecessary-pass - This can hurt readability
 # too-many-instance-attributes - This should be refactored later
 # duplicate-code - This should be refactored later as architecture has redundant Home-assistant devices.
+# too-many-arguments
 disable=
-  unnecessary-pass,too-many-instance-attributes,duplicate-code
+  format,unnecessary-pass,too-many-instance-attributes,duplicate-code,too-many-arguments,too-many-boolean-expressions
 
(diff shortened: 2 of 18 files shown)


alandtse:
To test this, use this [branch](https://github.com/alandtse/home-assistant/tree/tesla_async) of HA. I will be submitting it after a few days testing.

Pipfile
```diff
@@ -12,6 +12,10 @@ pytest-cov = "*"
 tox = "*"
 twine = "*"
 python-semantic-release = "*"
+black = "*"
 
 [packages]
-requests = "*"
+aiohttp = "*"
+
+[pipenv]
```
zabuldon: Let's freeze versions.

pylintrc
```diff
@@ -1,10 +1,15 @@
+[MASTER]
+ignore=tests
```
zabuldon: Do we really want to ignore test in linting? In my opinion we have to keep them readable too.

setup.cfg
```diff
@@ -3,4 +3,56 @@ version_variable=teslajsonpy/__version__.py:__version__
...
+hvcs=github
+
+[pydocstyle]
+ignore = D202, D212, D416, D213, D203, D407
+
+[flake8]
+exclude = .venv,.git,.tox,docs,venv,bin,lib,deps,build
+# To work with Black
+max-line-length = 88
+# E501: line too long
```
zabuldon: Let's increase max-line-length to 120 instead of ignoring E501

alandtse: I'm just taking the HA ignores for now. They default to 88 since they're using black.

zabuldon: We don't need to keep HA code-style in our repo as long as it separated library.

alandtse: True, but remember one of the goals #24 was to potentially move some of the HA-specific code over to HA? It'd make it easier if we enforced similar code-styles.

I'm pretty sure black would ignore the 120 length anyway and bring it back down to 88.

zabuldon: @alandtse ok.

pylintrc
alandtse: Ok, removing the ignore.

Pipfile
alandtse: Can't we rely on the Pipfile.lock file?

alandtse:
Whenever you're ready to merge this the semantic-release workflow should kick in. It's currently set with a `--noop` flag so we can confirm it thinks the next release is v1.0.0.  We'll need to remove that --noop flag to test the github upload and pypi release. I believe all those credentials should work as the debug is showing they aren't blank (they're masked though).

PR #57, opened 2020-01-28
feat: add websocket support

alandtse:
Needs following update to [HA](https://github.com/home-assistant/home-assistant/compare/dev...alandtse:websocket?expand=1) to test.

Apparently Tesla only allows one app to consume websockets at a time so may block other apps from gathering location info. This is now enabled as an integration option.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,7 @@ black = "*"
 
 [packages]
 aiohttp = "*"
+backoff = "*"
 
 [pipenv]
 allow_prereleases = true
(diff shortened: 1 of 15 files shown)


sillyfrog:
I have been testing this for a few days now and it's looking great (early Model S). As a thought in light of your comment about only one websocket connection allowed at a time, should this feature be optionally turned off in the configuration (for those users that may not want it)?

PS: If I should put this in an issue, please let me know.

Thanks!

PR #61, opened 2020-03-05
feat: add sentry mode switch to fix issue #43

hobbe:
Add a switch to enable/disable sentry mode for Tesla vehicle in Home Assistant. This should fix issue #43.

PR #66, opened 2020-03-17
fix: remove checks for state for sentry mode

hobbe:
Similar to PR #64, and as requested in PR #61, remove check for current state in case it is not accurate.

### DIFF
--- a/teslajsonpy/sentry_mode.py
+++ b/teslajsonpy/sentry_mode.py
@@ -44,21 +44,21 @@ class SentryModeSwitch(VehicleDevice):
         last_update = self._controller.get_last_update_time(self._id)
         if last_update >= self.__manual_update_time:
             data = self._controller.get_state_params(self._id)
-            self.__sentry_mode_available = data and data["sentry_mode_available"]
-            if self.__sentry_mode_available:
-                self.__sentry_mode = data["sentry_mode"]
+            if data and "sentry_mode_available" in data:
+                self.__sentry_mode_available = data["sentry_mode_available"]
+                if self.__sentry_mode_available and "sentry_mode" in data:
+                    self.__sentry_mode = data["sentry_mode"]
+            else:
+                self.__sentry_mode_available = False
+                self.__sentry_mode = False
 
-    def is_available(self):
+    def available(self):
         """Return whether the sentry mode is available."""
         return self.__sentry_mode_available
 
-    def is_enabled(self):
+    def is_on(self):
         """Return whether the sentry mode is enabled, always False if sentry mode is not available."""
         return self.__sentry_mode_available and self.__sentry_mode
-
-    def get_value(self):
-        """Return whether the sentry mode is enabled."""
-        return self.is_enabled()
 
     @staticmethod
     def has_battery():
--- a/teslajsonpy/sentry_mode.py
+++ b/teslajsonpy/sentry_mode.py
@@ -56,20 +56,18 @@ class SentryModeSwitch(VehicleDevice):
 
     async def enable_sentry_mode(self):
         """Enable the sentry mode."""
-        if not self.__sentry_mode:
-            data = await self._controller.command(
-                self._id, "set_sentry_mode", {"on": True}, wake_if_asleep=True
-            )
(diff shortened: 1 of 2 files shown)


alandtse:
Given the [request](https://github.com/home-assistant/core/pull/32938) in the main HA, I'll leave this open for any further fixes we can batch together.

hobbe:
Regarding the issue in HA PR https://github.com/home-assistant/core/pull/32938, we only know that sentry mode is available when we query `vehicle_state`, that is when `update()` is called.
This means that update needs to be called before any setup in HA.

alandtse:
I think you can just set the [`entity_registry_enabled_default`](https://developers.home-assistant.io/docs/entity_index#advanced-properties) parameter to `False` on init and then enable it once the car is queried and we know.  Probably makes sense to add that parameter to vehicle.py as `True` for default. In sentry_mode.py you can then add the KeyError check on async_update.

hobbe:
@alandtse , I have added a flag to know if sentry mode is available. If not, nothing should happen and sentry mode is always False.

As long as there is no flag at the vehicle level, then the switch will exist but do nothing. Can a switch be set to disabled?

teslajsonpy/sentry_mode.py
```diff
@@ -43,11 +44,21 @@ def __init__(self, data, controller):
         last_update = self._controller.get_last_update_time(self._id)
         if last_update >= self.__manual_update_time:
             data = self._controller.get_state_params(self._id)
-            self.__sentry_mode = data and data["sentry_mode"]
+            self.__sentry_mode_available = data and data["sentry_mode_available"]
+            if self.__sentry_mode_available:
+                self.__sentry_mode = data["sentry_mode"]
+
+    def is_available(self):
```
alandtse: Change this to `available` so [HA can use it directly](https://developers.home-assistant.io/docs/entity_index#generic-properties).

teslajsonpy/sentry_mode.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, data, controller):
         """
         super().__init__(data, controller)
         self.__manual_update_time = 0
+        self.__sentry_mode_available = False
```
alandtse: Isn't this unnecessary since it's in vehicle?

teslajsonpy/sentry_mode.py
```diff
@@ -43,11 +44,21 @@ def __init__(self, data, controller):
...
+            self.__sentry_mode_available = data and data["sentry_mode_available"]
+            if self.__sentry_mode_available:
+                self.__sentry_mode = data["sentry_mode"]
+
+    def is_available(self):
+        """Return whether the sentry mode is available."""
+        return self.__sentry_mode_available
+
+    def is_enabled(self):
+        """Return whether the sentry mode is enabled, always False if sentry mode is not available."""
```
alandtse: HA wants you to return `None` typically for values, but I'm not sure for switches. Can you test HA with `None` to see what the behavior is for a switch?

hobbe: OK, I will test this.

hobbe: A None value behaves as a False. Change to `available` and `is_on as per HA`entity.

teslajsonpy/sentry_mode.py
hobbe: Changed to `available`.

hobbe:
Argh... I am too late. I removed the unecessary `__sentry_mode_available`, as it is already available in VehicleDevice as `sentry_mode_available`. I keep it for next PR.

teslajsonpy/sentry_mode.py
hobbe: Correct, it is in VehicleDevice as `sentry_mode_available`. `__sentry_mode_available` can be removed.

alandtse:
Ooops. I thought you were done since you addressed everything. If you know anything about using pytest, it'd be nice to have a suite of tests for sentry mode.  That may be a worthwhile PR to add and then we can remove it in that one too.

PR #67, opened 2020-03-21
feat: add sentry mode availability flag to vehicle

hobbe:
As a follow-up of PR #66, this PR adds a sentry mode availability flag to the Vehicle.

The purpose of the flag is to not create a SentryModeSwitch in Home Assistant if the vehicle does not support sentry mode.

See also https://github.com/home-assistant/core/pull/32938#discussion_r394620506


alandtse:
Let's move this into #66 since the CI will bump on every PR and this is related anyway.

hobbe:
Moved to PR #66.

Issue #50, opened 2019-12-03
Support unlock trunk/frunk


phaza:
https://tesla-api.timdorr.com/vehicle/commands/trunk

Would be really nice!
I use Apple Watch integration with the HA Companion app, and would love to just tap a button on the watch to unlock the trunk/frunk.

cby016:
+1 for this feature

alandtse:
Closed by #71

PR #71, opened 2020-03-28
feat: add trunk and frunk management

hobbe:
In order to fix issue #50, this PR adds:

- 🔒 lock for rear trunk (lock/unlock)
- 🔒 lock for front trunk (frunk) (lock/unlock)


alandtse:
Aren't the sensors redundant of the switch?

hobbe:
Yeah. I needed to know if the trunk was open. Then I thought a switch would also enable opening/ closing it.
But in the end, I think a lock would be more appropriate. 
I will do some cleaning.

hobbe:
This PR adds trunk/frunk locks.
No changes are necessary in HA, the locks are added out-of-the-box :+1:

alandtse:
Once this is merged in you will need to do a HA PR to do the version bump and explain the new features it brings in. Also you'll need to do the documentation update.

PR #82, opened 2020-06-07
Added window lock

miguelangel-nubla:


alandtse:
Thanks for the contribution! Let me test it and get back to you on any feedback.

alandtse:
Thanks. It's working on my system.  Should this really be a lock? I feel like this should be a switch. We can probably call it Window vent. What does the app call it?

miguelangel-nubla:
The apps shows it as "vent" right next to the door locks and the GUI layout suggest they are of the same "kind".

The main reason I think of it as a lock is that generally, in home automation systems, locks are treated differently with an added layer of security and/or confirmation from the user.

And conceptually I also don't think would be right to allow the car state to be "locked" while a simple switch for the windows is off effectively making it, at least partially, "unlocked".

alandtse:
I'm not sure I buy that it should be a lock in that if I think about a window lock I think of a child safety lock where it can't be opened until unlocked. @hobbe any thoughts on whether the vent mode should be a lock?

hobbe:
> any thoughts on whether the vent mode should be a lock?

It is difficult to represent the state of the window with a [switch](https://www.home-assistant.io/integrations/switch/) (actions: turn_on, turn_off, toggle).
When ON, is the window open or closed (car is secured)?

And I agree that it is strange for a window to appear as a [lock](https://www.home-assistant.io/integrations/lock/) (actions: lock, unlock, open).
But it could fit. I chose this solution for the trunk.

For the windows, maybe the [cover entity](https://www.home-assistant.io/integrations/cover/) is more appropriate (actions: open_cover, close_cover, stop_cover)

> The following device classes are supported for covers:
window: Control of a physical window that opens and closes or may tilt.

miguelangel-nubla:
I agree that a lock is not the perfect fit for it. Cover entity is definitely the entity home assistant has designed for this.

Personally I would have to add a extra layer of security as those entities are exposed as is and would not like to be able to open them by accident, but I understand that would need to be part of my personal implementation. Maybe even push for a "secure" attribute on the cover entity...

alandtse:
Thanks all for the discussion. Let's go with a cover. I also asked in the HA discord and people are saying it's a cover. @miguelangel-nubla can you convert it to a cover please?

miguelangel-nubla:
Done!

You will need a modified tesla component for home assistant in order to load the new cover entity: https://github.com/miguelangel-nubla/hass_tesla

The problem I see now is the buttons on home assistant to open and close are exactly the reverse of a car window.
Anyone knows a way to fix that without a custom card?
Capture

teslajsonpy/window_cover.py
```diff
@@ -0,0 +1,93 @@
...
+        self.__manual_update_time = 0
+        self.__closed_state = None
+
+        self.type = "window cover"
+        self.hass_type = "cover"
+
+        self.name = self._name()
+
+        self.uniq_name = self._uniq_name()
+        self.bin_type = 0x7
```
alandtse, line 45: You can remove bin_type. I don't think we actually use it for anything.

alandtse:
Nice, this works for me. One minor request to add typing hints.  

On the HA side is to make sure it handles if a car isn't reporting any data because it's asleep. You should set the state to unavailable if the `closed_state` is None. Right now it's reporting my device as open. There are other components that we probably need to fix to also properly report that unavailable state too, but it'd be good to have the cover start out correct.

On flipping the arrows, I asked in the HA dev discord and haven't gotten a response yet. I suspect we may have to ask HA for a property to flip open and close. Taking a quick look, it looks like [Zwave](invert_openclose_buttons) implemented it for their covers at one point. See also this [PR](https://github.com/home-assistant/core/pull/6832). Perhaps that code can be borrowed.

miguelangel-nubla:
I will add typing hints and take a look at the unavailable state.

That [PR](https://github.com/home-assistant/core/pull/6832) is intended when the open and close services trigger the reverse button press. In our case that is working correctly `cover.close_cover` and `cover.open_cover` do what they should do, the issue is with the GUI assuming the open service corresponds to an "UP" movement.

Let me know the solution you get from HA dev discord.

alandtse:
No response from the backend dev channel. I'll try the front end. This may not be resolved unless users complain.  It might be worth opening an issue with the frontend team to be able to specify it as an option.

miguelangel-nubla:
Thanks for the update.

Coming back to the unavailable state, I have been experimenting and the front end disables the buttons when the entity is unavailable. Do we really want that?
I mean, the way it is now the entity shows the last known cached state as with the other entities, and when you click on open or close it wakes it up in the background and runs the command. You can't do that with a unavailable entity.
I guess in your case it reported the cover as opened by default because it did not have data yet.
I propose to make it unavailable by default but keep it as always available after data is received (like it is now) so you can always act on it.

alandtse:
Good points about caching and making it interactable. If every button went unavailable, then users couldn't just wake up the device by changing a setting. I agree with your proposal that on the start we should set unavailable and then cache afterwards.  

One thing to note, we have to be careful about implicitly assuming that the vehicle can be woken up. If it's in a place without mobile access or perhaps mobile access is disabled because it's in service, it'll never wake up so even if we issue the command, it may never act. We have to also consider that case.

There is also an option in HA for [assumed states](https://developers.home-assistant.io/docs/core/entity#generic-properties) to indicate that the information may no longer be accurate. We should be setting that to true if we're reliant on older data. 

So in summary, I think we can probably approve this PR once the typing is put in. Afterward, you'll make the HA request with the proposal of starting unavailable and then updating and caching afterward.

alandtse:
One suggestion on flipping the UI is using device_class for the cover.

> usually characterization of platform entity types are done via the device_class property. Maybe that can be used and extended for this case? "car_window"?

Can you look into implementing that for this cover? The next step is to have the UI team know to read that property for covers. I don't work on the front end at all, but you could open an issue after you submit your HA PR with that.

miguelangel-nubla:
Setting the device class to "car_window" seems to behave the same for me on HA.
I don't know if it is a caching thing where it comes up as the latest *valid* device class for the entity automatically or it is working as expected.

alandtse:
I think if you don't have a device_class it ignores it and defaults to something. baloob said there may already be a cover device_class that has the inverted arrows. Can you check please?

teslajsonpy/window_cover.py
```diff
@@ -0,0 +1,91 @@
...
+        controller : teslajsonpy.Controller
+            The controller that controls updates to the Tesla API.
+
+        Returns
+        -------
+        None
+
+        """
+        super().__init__(data, controller)
+        self.__manual_update_time = 0
```
alandtse, line 35: Also please get the typing info for instance variables.
```python
                        self.__manual_update_time: int = 0
```

alandtse:
See https://github.com/home-assistant/core/blob/dev/homeassistant/components/cover/__init__.py#L43 and https://developers.home-assistant.io/docs/core/entity/cover/.

miguelangel-nubla:
> See https://github.com/home-assistant/core/blob/dev/homeassistant/components/cover/__init__.py#L43 and https://developers.home-assistant.io/docs/core/entity/cover/.

I don't know what are you meaning to say. I have been looking at that documentation before but I can't seem to find the "inverted arrows" device class.

alandtse:
> I don't know what are you meaning to say. I have been looking at that documentation before but I can't seem to find the "inverted arrows" device class.

Sorry, I meant for you to try each of the device classes to see if they already have one that inverts the arrow. If none work, then we will need to ask HA to add a device_class for car_window and that will be another issue/PR request you will need to submit. Even if  it doesn't exist yet, we can still merge this teslajsonpy PR in and you can submit your HA PR request. It'll just have that wierd UI bug until they accept a new `car_window` device class.

teslajsonpy/window_cover.py
```diff
@@ -0,0 +1,93 @@
...
+from teslajsonpy.vehicle import VehicleDevice
+
+
+class WindowCover(VehicleDevice):
+    """Home-assistant cover class for the windows of Tesla vehicles.
+
+    This is intended to be partially inherited by a Home-Assitant entity.
+    """
+
+    def __init__(self, data: dict, controller: 'teslajsonpy.Controller') -> None:  # noqa: F821
```
alandtse, line 19: Can't we just import the class above instead of suppressing the warning?

miguelangel-nubla: teslajsonpy/window_cover.py:1:0: R0401: Cyclic import (teslajsonpy.controller -> teslajsonpy.window_cover) (cyclic-import)

alandtse: Oh ok.

alandtse:
@miguelangel-nubla just wanted to see if you're still interested in this. I recognize there's a lot of steps which can be discouraging but HA has been tightening it's submission process. I've recently been working on some PRs to make it slightly easier for the component.

miguelangel-nubla:
Yes, hopefully I can dig into it next week.

millerkyle72:
I apologize if this is not the right place, but how do you modify the Tesla component in HA to include the include the window button? Do I just clone this (https://github.com/miguelangel-nubla/hass_tesla) in my custom_components folder? Not sure if there is more that I would have to do.

Great work on adding the window feature, I can't wait to automate my windows!

alandtse:
@millerkyle72 Yes, but it must be in the `custom_components/tesla` directory. Please note, this PR is not up to date with the main component anymore so you may not have all the same bug fixes we've put in the main component.

millerkyle72:
@alandtse Is it any easier to add this now that the Tesla integration is going through HACS?

alandtse:
> @alandtse Is it any easier to add this now that the Tesla integration is going through HACS? 

No. It's not updated for deployment nor in the HACS version.

alandtse:
Closing since this is stale and given the rewrite.

PR #85, opened 2020-06-24
Add option for filtering cars by VIN

### LINKED ISSUE #84: Filter cars returned to Home Assistant
Give the choice to limit the resulting components to a subset of the cars connected to the account.
This is very useful in case the user has access to more than one car but does not want to control all of them in Home Assistant. 

This could be achieved by passing an optional list of VINs to connect(...)

alexpilotti:
Closes #84

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -250,13 +250,14 @@ class Controller:
         self.enable_websocket = enable_websocket
 
     async def connect(
-        self, test_login=False, wake_if_asleep=False
+        self, test_login=False, wake_if_asleep=False, filtered_vins=None
     ) -> Tuple[Text, Text]:
         """Connect controller to Tesla.
 
         Args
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. Does this pass the `make lint`? I'm not sure it'll accept `VINs` given the capitalization. Let's change it to `filtered_vins` so it's clear.

I assume you've also confirmed that it's not trying to update those vehicles too right?

alexpilotti:
@alandtse done, thanks! Here's the successful make lint output: http://paste.openstack.org/show/795249/

alexpilotti:
> I assume you've also confirmed that it's not trying to update those vehicles too right?

Yes, only the entities belonging to the filtered vehicles are meant to be visible in Home Assistant

alandtse:
Thanks. I'll do a master build now so you can get a version # you can use to update HA.

alexpilotti:
Perfect thanks! I’ll get the HA patch ready in the meantime.

PR #117, opened 2020-10-18
Add charge limit soc

### LINKED ISSUE #114: Feature request: Soc limit.
The api seems to expose more info then we are using atm. I would like to have a soc_limit as a sensors or attribute in home assistant. If this is acceptable i can create a PR for this. This would be usefull you you want to setup your own smart charging.

Hellowlol:
The goal for this PR is to make charger_soc limit at a attribute in the ``sensor.tesla_model_3_battery_sensor``, I'm not sure if this is al that is required or if additial changes is needed. I got lost in all the abstractions of the api.

Closes #114

### DIFF
--- a/teslajsonpy/homeassistant/battery_sensor.py
+++ b/teslajsonpy/homeassistant/battery_sensor.py
@@ -26,6 +26,7 @@ class Battery(VehicleDevice):
         self.__battery_level: int = None
         self.__charging_state: bool = None
         self.__charge_port_door_open: bool = None
+        self.__charge_limit_soc = int = None
         self.type: Text = "battery sensor"
         self.measurement: Text = "%"
         self.hass_type: Text = "sensor"
@@ -49,6 +50,7 @@ class Battery(VehicleDevice):
         if data:
             self.__battery_level = data["battery_level"]
             self.__charging_state = data["charging_state"] == "Charging"
+            self.__charge_limit_soc = data["charge_limit_soc"]
 
     @staticmethod
     def has_battery() -> bool:


teslajsonpy/homeassistant/battery_sensor.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, data: Dict, controller) -> None:
         self.__battery_level: int = None
         self.__charging_state: bool = None
         self.__charge_port_door_open: bool = None
+        self.__charge_limit_soc = int = None
```
alandtse, line 29: This isn't correct syntax for typing. Should be:
```python
        self.__charge_limit_soc: int = None
```

alandtse (commented):
Thank you. While you are storing the value properly in the battery sensor you also need to do some other stuff for this to work.
1. Expose the self.__charge_limit_soc value through the API as a [property](https://github.com/zabuldon/teslajsonpy/blob/dev/teslajsonpy/homeassistant/charger.py#L244,L247). It's bad form to directly access an object variable prefaced with `__` so right now you don't have any mechanism to access the value outside. 
2. Modify the [device attributes](https://github.com/home-assistant/core/blob/dev/homeassistant/components/tesla/sensor.py#L84,L98) to expose something for battery sensors using what you have in 1 above. 
3. Write a [test](https://github.com/zabuldon/teslajsonpy/blob/dev/tests/unit_tests/homeassistant/test_battery_sensor.py) to cover your new property.

Hellowlol:
Thanks! I’ll fix that

alandtse:
Oh, for #2 above, you could also set the attr directly in teslajsonpy. See the [parking brake sensor](https://github.com/zabuldon/teslajsonpy/blob/dev/teslajsonpy/homeassistant/binary_sensor.py#L109,L111) for how that works. Right now there's a bug in HA where it wasn't automatically importing it. After that [PR](https://github.com/home-assistant/core/pull/42257) is merged, then you won't have to make any changes to HA itself.

alandtse:
By the way, what does charge_limit_soc actually represent?

Hellowlol:
Thanks for the examples. I’m gonna close this and open a new one. As I think this should be under the charger sensor and not battery

It’s the percentage where the car will stop the charger based on state of charge. Without this I have no way of knowing how many kWh I can add to the battery. As I never charge to 100%

Issue #166, opened 2021-03-09
Update-frequency while connected to a charger


Olen:
Today, the polling interval will be set back to default (11 minutes) when the car moves from "Charging" to e.g "Complete", even if the car is still connected to the charger.
I don't know if this is intentional, but personally, I would prefer to keep on polling every minute even if the charging is done as long as it is still _connected_ to  charger.

https://github.com/zabuldon/teslajsonpy/blob/5889b896dc20c11b4b2342cdd4af7963f6ad24e8/teslajsonpy/controller.py#L628

I don't know the best way to check if the car is still connected to the charger, as it has been some years since I played with the api, and remember from "back then" that some values could be a bit unreliable, and was lingering even after they should have been reset.
Some ideas:
- add an "or charging_state == 'Complete' (and maybe also 'Starting to charge" or whatever it is called) 
- add a check for 'charge_port_latch' == 'Enaged' 
- add a check for 'charger_pilot_current' > 0
- add a check for 'conn_charge_cable' is not None

If the drop in poll rate is intended, I would like to add something like a "always_poll_on_charger_connected" flag, so you can keep the default of not polling, but the user kan increase polling frequency if he wants.

alandtse:
It is intentional to allow the car to sleep since it is no longer charging. Otherwise it will unnecessarily drain the battery which has real world costs.

Olen:
Thanks.

I wanted to check if there was any chance of it being implemented before doing the job.
But I'll see if I can add a PR.  I'll probably implement it as a flag, so you can set something like 
"wake_up_policy:" 
- "always"  (just keep the car awake, no matter what)
- "connected" (keep it awake while connected to a charger, even if charging is completed)
- "charging" (current implementation, and default)

Does this sound like a reasonable approach?

alandtse:
Yes. But I don't think always is a good idea. You can select multiple flags and just populate the default and your setting.

Please note the latest dev branch is built to be used with the current [PR](https://github.com/home-assistant/core/pull/46558) to HA which is in testing.

Olen:
I have been running "always" forever, due to another 3. party app that has not implemented any sleep mode at all, and it has never caused any problem.  Whether it is a good idea or not, and how much it will cost you in terms of money and lost range while parked, is of course up to each to decide.

I'll add the PR here first, and it should be fairly easy to add it to the tesla integration in HA after that.

Olen:
Closing this as it has been fixed in the latest merge.

PR #167, opened 2021-03-12
Polling policy

Olen:
See #166 

This PR adds a new flag; "polling_policy" to allow for some granularity on the polling.
If "polling_policy" is set to "always", the controller will keep polling every "update_interval" except when the car is driving (when DRIVING_INTERVAL is used instead).
Be aware that this will probably never allow the car to sleep, so it might drain the battery more than necessary.

If "polling_policy" us set to "connected", the poller will run normally as long as the car is connected to a charger.  This will increase vampire drain slightly, but only while connected to a charger, so it should not cause a problem while parked without a charger connected.

The normal polling policy will still be default, and only poll while the car is driving, charging, or while sentry mode or climate is on.


alandtse:
The code looks fine. Do you mind adding tests just for this new parameter? I realize we don't have a test yet for the controller but we've been trying to improve on that.

Olen:
Hi, and thanks for the review.
I am not sure exactly where I should add that test.
It seems like all the tests in test/home-assitant/* are just running `_controller = Controller(None)` so they don't test any parameters at all, so I don't really know where to start...

To actually validate that it works as intended would be pretty complex, and I don't know the code well enough to write a usable test, I am afraid.
You would pobably need to first set up a default controller, force charge_state to "Charging", check the reslt from _calculate_next_interval.
Then force charge_state to "Complete" and then check that result from _calculate_next_interval
Then force charge_state to "Disconnected" and then check that result from _calculate_next_interval

And then do the same 3 tests again for polling_policy "connected" and "always".

alandtse:
I can probably set up a framework file with one example if you'd be willing to do the rest. All you have to do is mock the API state and then test the `_calculate_next_interval` function to see which time it is returning.

Olen:
Thanks. Please set up an example, and I will do the rest.

Olen:
Hm,
I thought I could just add some more tests here:
https://github.com/zabuldon/teslajsonpy/blob/dev/tests/unit_tests/homeassistant/test_charger_connection_sensor.py

But `_calculate_next_interval()` is a nested function inside `async def update()`, and apparently nested functions can't really be tested very easily...

alandtse:
It should probably be in test_controller.py (not created yet) and you're right you can't test inner functions.  We'll have to mock the API responses (like the other test functions) to get it into the different states.

Olen:
Yeah. Getting it into the states is one thing. But I am trying to wrap my head around how to test the "next interval", since it is only used in one location, and only to check if it should append a task.

An idea I am playing with is to monitor some arbitrary value (like "battery_level")
- Set charging state to Charging
- run update()
- Verify that battery_level is the default value (64)
- set battery_level = 10 somehow
- Run "update()
- Verfy that it is still 64 (should not have updated the value, since it is less than 1 minute since the last update)
- Force _last_update_time[vin] to be more than one minute ago  
- Run "update()
- Verify that battery_level is now the mocked value (10)


And then do the same for different states and last_update_time to validate if it actually runs an update or not.

alandtse:
Thinking about it more, let's break it out as a helper function and write a test against it.

alandtse:
> When you say a helper-function is it as easy as moving the `def _calculate_next_interval` up a level, and out of the update() function?

Yup, that's all I meant.

Olen:
So many private variables in the controller...
Had to add quite a few other functions to be able to test all the settings.  But I believe this should work now.

Issue #186, opened 2021-04-03
KeyError: 'climate_state'


alandtse:
@Trupal00p 

I'm getting this error in a live system. Can you please check and provide a fix? Please note there will be vehicles that may not have seat heaters.

```
  File "/workspaces/home-assistant/homeassistant/components/tesla/config_flow.py", line 229, in validate_input
    config = await controller.connect(
  File "/workspaces/home-assistant/config/teslajsonpy/controller.py", line 316, in connect
    self._add_components(car)
  File "/workspaces/home-assistant/config/teslajsonpy/controller.py", line 549, in _add_components
    self.__components.append(HeatedSeatSwitch(car, self, 'left'))
  File "/workspaces/home-assistant/config/teslajsonpy/homeassistant/heated_seats.py", line 47, in __init__
    self.__seat_heat_level = data['climate_state'][f'seat_heater_{seat_name}']
KeyError: 'climate_state'
```

I'll spawn a new issue with this comment so please respond there.

_Originally posted by @alandtse in https://github.com/zabuldon/teslajsonpy/issues/180#issuecomment-812803500_

alandtse:
I will put a temporary fix that won't load the seat sensors if a keyerror is found. That may be the solution, but I'd appreciate you taking a look and maybe considering another fix with a test for this case.

alandtse:
It looks like you're relying on the sensors to be enabled. However, the data won't have been pulled yet during init. You should move that into the update function.

Trupal00p:
Just opened a PR to fix this.  Thanks for clarifying what was causing the issue, misunderstood the lifecycle of the tesla data when I originally added.

PR #189, opened 2021-04-06
Fixes for initialization of heated seats

Trupal00p:
Fixes for #186


alandtse:
Can you please add a test for the case where it broke before where the data isn't initialized yet?

alandtse:
We're close to pushing a new release to address the login errors. Any chance to get that test in place so I can merge this too?

Trupal00p:
Sorry for the delayed reply.  I think the changes I made on 47e5f23 in test_heated_seat.py (line 33) address this.  It's the same approach used by the initialization tests for the other components.

alandtse:
I mean do we have a test in place to make sure we don't get the exception on startup #186 ? In theory after your fixes, you won't get the exception but this is to prevent a regression in the future.

alandtse:
Ok. Thanks!

PR #206, opened 2021-08-09
Fix bug in refresh_access_token

BreakingBread0:
Throws an exception when calling refresh_access_token that the URL parameter is nether a string nor a httpx.Url

### DIFF
--- a/teslajsonpy/connection.py
+++ b/teslajsonpy/connection.py
@@ -553,7 +553,7 @@ class Connection:
             "scope": "openid email offline_access",
         }
         auth = await self.websession.post(
-            self.auth_domain.with_path("/oauth2/v3/token"),
+            str(self.auth_domain.with_path("/oauth2/v3/token")),
             data=oauth,
         )
         return auth.json()


alandtse:
Is this the only fix you need in teslajsonpy or are you expecting more?

alandtse:
I pushed to dev but will wait to pushing for publication for now. If you need multiple changes I'll wait till the set is there.

BreakingBread0:
> I pushed to dev but will wait to pushing for publication for now. If you need multiple changes I'll wait till the set is there.

Oh, thank you. I actually have another issue with the library:
https://github.com/zabuldon/teslajsonpy/blob/ba19cc1d3182979d6ae49bb7606bf74853d586db/teslajsonpy/connection.py#L118
Is this check for refresh_token AND checking for expiry intentional? Shouldn't only the access_tokens (for ownerapi and sso) expire?

alandtse:
Yes only access tokens expire but if you do not have a refresh token, you can't get an access token. So it's confirming there is a valid refresh token before trying to refresh.

Issue #192, opened 2021-04-27
Seat heater architecture plans


mattsch:
After poking around a bit at the new seat heater functionality (thanks @alandtse and @Trupal00p), I starting thinking about current behavior and how it will integrate into HA.

Firstly, the current approach of checking for availability at startup can't work since it doesn't have the data for the car until after the initial setup and then **only** if the car is awake. My thought is that it should either add those as disabled entities by default and/or have a config option in the integration to enable them. I have a proof of concept on a local branch for the former option.

The other (probably more controversial option) is that we shouldn't actually represent them as switches but as sensors and then expose a service (tesla.set_heated_seats?) that controls them. There's no entity type that accurately represents the functionality so anything we do will be a hack. Fans might work as long as they can only have the `SUPPORT_PRESET_MODE` feature, and climate could work if we only set the `SUPPORT_PRESET_MODE` feature, but either one is a bit dirty IMO. I'm happy to work on an implementation for the service if that would be helpful as well.

alandtse:
Do any cars have a seat warmer entity? Should we propose an entity for that?

While it's fine to just choose something, we have the opportunity to lead on this. Perhaps we need to reach out to the other car integrations.

On a general service, I'm tempted to opening a service call that allows accessing everything in the api. Someone on the forum wants to adjust the standard charging limit and that would require such a service.

mattsch:
I haven't seen any mention of it, I'm sure other cars do but I don't see any other integrations that expose the functionality. I do like how the [BMW](https://www.home-assistant.io/integrations/bmw_connected_drive/#services) integration has services for the various functionality. Makes it very easy to see what it can do, and less error-prone on the user side. If we followed that same approach we could expose quite a bit more of the native commands that also don't have good translations to entity types.

If we wanted to go with entities, there is the [number](https://developers.home-assistant.io/docs/core/entity/number/) entity that might be a solid enough option and would work for a charging limit entity.

alandtse:
I wouldn't want to go too far off the trail but to talk the entity approach through. Cn we tie the number entity to the max range switch? For example, any number besides 90% changes the command to the argument based setting?


If BMW is doing the exposed services, maybe we should too. We could make it generic so if you know the endpoint and the command, we pass it through without further checking. Do you want to take a shot at that?

mattsch:
[...]

PR #210, opened 2021-08-18
Add Heating Steering Wheel Support & Move Heated Seats to SELECT

megabytemb:
Hello, 

I don't usually add to Open Source, so please go easy :)

As a pre-cursor to adding heated Steering Wheel support into Home Assistant, I've added the steering wheel support into this library.

Also, Heated seats, although in the library, was never implemented in Home Assistant.
Home Assistant now supports "Select" entities, which works really nicely for heated seats.

If you approve this Pull request, i'll then do another for Home Assistant directly to add this functionality.

Added Tests and confirmed linting as required.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -13,3 +13,4 @@
 -   ultratoto14 [GitHub](https://github.com/ultratoto14)
 -   johanjongsma [GitHub](https://github.com/johanjongsma)
 -   hobbe [GitHub](https://github.com/hobbe)
+-   megabytemb [GitHub](https://github.com/megabytemb)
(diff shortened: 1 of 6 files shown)


alandtse:
Thanks for doing this. Sorry I didn't even realize you had submitted this as my notifications are off here.

alandtse:
Would this close #192 too?

megabytemb:
I wasn't aware of that conversation going on. 

I believe it does. Select is a new entity type in Home Assistant, and IMO, fits the use case of a seat warmer perfectly.

However, Perhaps we should look at how to handle "errors" when the climate control isn't turned on. At the moment, it just silently fails.
I think just having the feature in home assistant, and appropriate documentation around it is good enough, but always open for discussion.

alandtse:
Thanks. We currently are silent for all errors unless it's an underlying library exception. But we probably could throw an exception. I'm ok with this for now and will add it to dev. Please consider adding the exception throwing in another pr.

Issue #205, opened 2021-08-08
Captcha required on sign-in?


k-korn:
Environment: latest Hass with teslajsonpy 0.18.3 
Issue: unable to authenticate with Tesla api, response to https://auth.tesla.com/oauth2/v3/authorize contains "Enter the characters in the picture" type captcha.

Up until recently, Tesla integration for HomeAssistant worked flawlessly.
Now it shows car entities as "unavailable", and if I add "teslajsonpy: debug" into Hass logger, the following output is produced:
 (possibly identifying data replaced with "...")

```
Aug 08 12:06:54 hass[26844]: 2021-08-08 12:06:54 DEBUG (MainThread) [teslajsonpy.connection] Attempt #0
Aug 08 12:06:54 hass[26844]: 2021-08-08 12:06:54 DEBUG (MainThread) [teslajsonpy.connection] POST:
Aug 08 12:06:54 hass[26844]: https://auth.tesla.com/oauth2/v3/authorize?client_id=ownerapi&code_challenge=N...NQ%3D%3D&code_challenge_method=S256&redirect_uri=https://auth.tesla.com/void/callback&response_type=code&scope=openid+email+offline_access&state=b1...MQ with
Aug 08 12:06:54 hass[26844]: Headers({'host': 'auth.tesla.com', 'accept': '*/*', 'accept-encoding': 'gzip, deflate', 'connection': 'keep-alive', 'user-agent': 'HomeAssistant/2021.8.2 httpx/0.18.2 Python/3.9', 'cookie': '_abck=2...D~-1~-1~-1; ak_bmsc=79...bQ==; bm_sv=58A3...3Dx0=; bm_sz=EEC0...5601; tesla-auth.sid=s%3Ahc...uqw', 'content-length': '150', 'content-type': 'application/x-www-form-urlencoded'})
[...]

PR #211, opened 2021-08-27
fix: allow auth using refresh_token

omelhus:
This fix will allow a user to provide a refresh token generated by apps like "AuthAppForTesla" etc to authenticate. This is the way Telsamate, Teslascope and others are authenticating and it works.

This will require that the core Tesla integration in Home Assistant is updated to be able to provide the refresh_token, but this change is a prerequisite.

Related issue #205.


alandtse:
Thanks. I know we don't really have testing on the user auth yet, but do you mind mocking a test for this new path?

omelhus:
No clue on how to mock a test for this, sorry.

Issue #49, opened 2019-11-11
Add support for Tesla Solar


JoshuaSG:
I was thinking about adding support for Tesla Solar, it would be nice to pull the current status of my solar system and then integrate it into Home Assistant.

It looks like you might be in the middle of refactoring, should I wait a bit before I start digging in, or go ahead and attempt something.

alandtse:
We should be refactored now to async if you want to start adding.

ocalvo:
I would love to see this implemented.
How can I help?

JoshuaSG:
Sorry I haven't had time to work on this, Python is a new language to me so I've been meaning to study up on a it. If you know Python I could point you in the right direction on the Tesla API endpoints.

hobbe:
In this context, there is a Tesla Powerwall integration coming to HA: https://github.com/home-assistant/core/pull/32851

alandtse:
I'll close this now that the [Powerwall integration](https://www.home-assistant.io/integrations/powerwall/) exists. That's a synchronous library, so if there was interest in merging it here, I wouldn't be opposed. But one reason it makes sense to be a different library is that manages to do  work locally while this component requires hitting Tesla's actual servers. I'm not sure if there's added benefit to have the connection to Tesla.

alandtse:
Ok, didn't realize there was a use case like that. I'll reopen. I don't have Tesla solar so while it shouldn't be hard to add the API calls, I wouldn't be able to test it and integrate it into HA. 

One place to look is at [tesla_api](https://github.com/mlowijs/tesla_api/blob/master/tesla_api/energy.py) which has mapped out the functions. You'd have to convert their calls to use our `connection.py` since that handles the login automatically.

Here is how I'd approach this.

1. Make sure the basic API calls exist using tesla_api as a guide. Right now we've actually embedded them in our controller component. The controller.py is the brains of the current architecture that controls the connection to the API to try to avoid overloading it.
2. Create solar specific entities to process the data collected from the controller. Right now, the component basically creates entities that are directly imported into HA. This is the confusing part for most people looking at this library in that we've basically moved all the intelligence from the HA component into this library.  All the current entities used in HA are part of the [vehicle entity](https://github.com/zabuldon/teslajsonpy/blob/dev/teslajsonpy/vehicle.py)
3. Modify the component in HA so it will load whatever you've generated in 2.

nomer:
I'm familiar with python as well as the Tesla API, but new to Home Assistant development. 

[...]

PR #212, opened 2021-08-28
Add Tesla Solar Energy Monitoring

bassrock:
This is an initial attempt at adding Tesla Solar energy to the library.

Before I go further, I'd like to get a once over of the PR and the approach. 

I have tesla solar working with HA via a Rest Sensor but would like to see it get more integrated with the offical plugin.


alandtse:
Thanks. Approach looks fine. Please remember to add tests. We've been trying to get better test coverage for the library since it's getting more complex.

alandtse:
Actually, one thing to note. One of the issues we have with this library is that we have so much embedded HA stuff down in the API. That makes it so if you want to make a change in HA, you have to make changes in 2 places: the api library and HA.  You'll note we have a long open issue #24 to try to break the HA out.

If we're starting from scratch here, I'd suggest you see if you can isolate the raw api calls so they're in this library and then build the entity logic in HA itself. So basically I wouldn't suggest keeping the homeassistant/* files. The downside of course is someone can't access entities in teslajsonpy, but I'm not sure who is actually using the library beyond HA. You may also want to see if we can make ourselves API compatible with the existing [powerwall](https://www.home-assistant.io/integrations/powerwall/) integration. That way you could just contribute to that integration directly too. I believe the main difference between that integration and what is happening here is the powerwall integration relies on local access while this would rely on cloud access. See #49

bassrock:
@alandtse yea I looked into local access unfortunately it doesn't have any. I have SolarCity panels with a Tesla gateway after the fact which seems to only use cloud.

I'll look at all those prs you mentioned. I also wasn't a big fan of having to do code here and then pr to the HA integration 😞

alandtse:
Yah, so if you're able to just make the necessary API changes here, you can take your HA specific code into HA itself. The only thing we'd need to keep in this component would be the base solar site object so the controller knows it should be fetching and caching information for that device.

alandtse:
If you're still working on this, I'd suggest seeing #216 to see if it simplifies your PR. I think that's a more robust way to define/access the underlying API. I'm planning to switch the vehicle related calls to that method and may also let us get the solar stuff for free.

Please also note, that with Tesla being removed from core, any improvements should go to the [custom_component](https://github.com/alandtse/tesla) I'm hosting now.

giachello:
Hi, did you end up merging this into the main branch. I have also started writing support for "energy_sites" (i.e. solar panels, not powerwalls).

bassrock:
@giachello i have not had time to continue this. go for it

PR #217, opened 2021-09-13
Add id_token to connect() response and return correct refresh_token

omelhus:
Hey,

This will return the id_token as well so that we can use PyJwt in the component to extract the email from the token directly.

```python
import jwt

[...]

def get_email_from_id_token(id_token: str):
    decoded = jwt.decode(id_token, options={"verify_signature": False, "verify_aud": False})
    return decoded["email"] if "email" in decoded else None
```

It also sets the refresh_token to the correct refresh token. The current version returns the refresh_token from the owner-api and that's not used for anything. 

See https://tesla-api.timdorr.com/api-basics/authentication#post-https-auth-tesla-com-oauth-2-v-3-token-1
> This uses the SSO refresh_token from Step 3 above to do an OAuth 2.0 Refresh Token Grant. This does not work with the refresh_token provided by the Owner API. Those have no use currently and should be discarded.

### DIFF
--- a/teslajsonpy/connection.py
+++ b/teslajsonpy/connection.py
@@ -63,6 +63,7 @@ class Connection:
         self.api: Text = "/api/1/"
         self.expiration: int = expiration
         self.access_token = access_token
+        self.id_token = None
         self.head = None
         self.refresh_token = refresh_token
         self.websession = websession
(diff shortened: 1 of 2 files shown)


alandtse:
Thanks. Is this a breaking change if we're changing the return signature?

omelhus:
No, it just returns an extra entry in the returned dictionary, so it's not a change in signature per say.

PR #218, opened 2021-09-14
feat: add id_token to connect() response and return correct refresh_t…

alandtse:
…oken (#217)

* Add id_token to connect() response

* Set refresh token to the one returned from auth-api and not owner api


alandtse:
Do you think this is the cause of any of the login errors people are still experiencing? We haven't updated the component to use this yet.

PR #219, opened 2021-10-09
Added support for Energy Sites 

giachello:
I added support for Energy Sites (Solar Panels). This has been working for my installation for a week now and feeds into my Energy panel on Home Assistant. I know you wanted to restructure the entire API in a separate thread on this topic, but I'm not going to do that work (don't have the time). 

This goes together with a pull request on the Tesla custom component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,6 @@
 
 <!--next-version-placeholder-->
 
-## v0.21.2 (2021-10-05)
-### Feature
-* Add State Class on Power Sensor
-
-## v0.21.1 (2021-10-02)
-### Feature
-* Add Support for Energy Sites
-
 ## v0.21.0 (2021-09-14)
 ### Feature
 * Add id_token to connect() response and return correct refresh_token ([#217](https://github.com/zabuldon/teslajsonpy/issues/217)) ([`6d837ed`](https://github.com/zabuldon/teslajsonpy/commit/6d837ed1b3a9794f4e9ffc3a0c59afc33702137e))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 <!--next-version-placeholder-->
 
+## v0.21.2 (2021-10-05)
+### Feature
+* Add State Class on Power Sensor
+
 ## v0.21.1 (2021-10-02)
 ### Feature
 * Add Support for Energy Sites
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 <!--next-version-placeholder-->
 
+## v0.21.2 (2021-10-05)
```
alandtse, line 5: No need to edit the version or changelog directly. We use semantic release so it will auto increment and fill out data.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "teslajsonpy"
-version = "0.21.0"
+version = "0.21.2"
```
alandtse, line 3: Revert the version change, semantic release will handle.

teslajsonpy/__version__.py
```diff
@@ -8,4 +8,4 @@
 https://github.com/zabuldon/teslajsonpy
 """
 
-__version__ = "0.21.0"
+__version__ = "0.21.2"
```
alandtse, line 11: Revert.

pyproject.toml
giachello: done

CHANGELOG.md
giachello: done

giachello:
Ok, I think I took care of all the comments you left. Let me know!

shred86:
Hey @giachello, I'm helping with the rewrite and had a question on some changes in this pull request (thanks for adding this by the way!). I was wondering what the thought was behind mapping the site_id to energysite_id?

````
        self.__id_energysiteid_map = {}
        self.__energysiteid_id_map = {}
````

It looks like all the power data is stored in `self.__power` which is organized by `energysite_id`, so even someone that has two energy sites, I can still just pull the data I need using `energysite_id`. I'm sure there's a good reason, I'm just a novice programmer so I'm trying to understand before I make any major changes. Thanks!

giachello:
Actually, I think I've just done that for completeness, to conform with the way that the Car entities were written at the time. Not sure if there is a real need to do this.

alandtse:
CarIDs are subject to change so it's possible they may change within a session so we had to translate it. If the energy sites cannot change, then you wouldn't need it.

shred86:
Ahh okay, that makes sense. I'm not 100% certain but I have not noticed the energy_site_id change. I haven't really looked closely at just the "id", but I'll keep an eye on it. Thanks!

PR #227, opened 2021-10-16
Fixed bug when energy site name is null

giachello:
closes https://github.com/alandtse/tesla/issues/62

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -48,7 +48,7 @@ from teslajsonpy.homeassistant.lock import ChargerLock, Lock
 from teslajsonpy.homeassistant.sentry_mode import SentryModeSwitch
 from teslajsonpy.homeassistant.trunk import FrunkLock, TrunkLock
 from teslajsonpy.homeassistant.heated_steering_wheel import HeatedSteeringWheelSwitch
-from teslajsonpy.homeassistant.power import PowerSensor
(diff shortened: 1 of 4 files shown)


teslajsonpy/controller.py
```diff
@@ -346,7 +346,10 @@ async def connect(
             energysite_id = energysite["energy_site_id"]
             self.__id_energysiteid_map[energysite["id"]] = energysite_id
             self.__energysiteid_id_map[energysite_id] = energysite["id"]
-            self.__energysite_name[energysite_id] = energysite["site_name"]
+            if "site_name" in energysite:
+                self.__energysite_name[energysite_id] = energysite["site_name"]
```
alandtse, line 350:
```python
                self.__energysite_name[energysite_id] = energysite.get("site_name", "My Home")
```

alandtse: Of course, please lint and make sure it works with a test.

alandtse (changes requested):

gwendalg:
The change is not enough: "site_name" is also used in homeassistant/power.py, line 37:
`self._display_name: Text = data["site_name"]`
Using .get() with a default value is preferable, as the `_display_name` field is actively used in homeassistant.

giachello:
Using get and fixed power.py as well. implemented tests.

PR #248, opened 2021-12-05
Expose max charger variable

danielp370:
Here is a PR to expose Charger's charge_current_request_max variable.  

I have a use case where I have multiple chargers (and cars) with different maximum charge capabilities. It would be really helpful to have max exposed so that I know up-front how much current I can push to each one by querying this variable.

Thanks


alandtse:
Thanks. This looks good. I'll merge it in so you can get it on pypi .

danielp370:
Thanks. Confirming it looks good in HA.

PR #252, opened 2021-12-05
Add Horn & FlashLight switch

rafal83:


alandtse:
Thanks. Please remember to include some basic tests and ensure the CI passes.

Also, wouldn't these be better as HA [buttons](https://developers.home-assistant.io/docs/core/entity/button/) instead of switches because they won't have on states. The difference is that if it's a button, you only have to implement the `async_press` method.

PR #259, opened 2021-12-11
fix(energy-sites): make name unique

bassrock:
When having 2 energy sites with no names, the `uniq_name` function needs to be unique for both to show up in HomeAssistant. To do this I added the site_id into the name.


alandtse:
Use a mock for testing. Also I thought I saw a email/token in that. Please delete. You may have to regenerate your tokens with Tesla to invalidate what you posted. No idea if GitHub can purge that data.

bassrock:
whoops. thank you @alandtse

PR #260, opened 2021-12-11
fix(energy-sites): make name unique

bassrock:
When having 2 energy sites with no names, the uniq_name function needs to be unique for both to show up in HomeAssistant. To do this I added the site_id into the name.

Redo of a previous pr that had secrets and no tests.


bassrock:
Note that about 10 minutes after I called `https://owner-api.teslamotors.com/api/1/energy_sites/:site_id/site_name` and added a site name, this integration started working and showing both sites without this change.

alandtse:
Does this cause a breaking change in HA if the site name is changing?

bassrock:
@alandtse yes it does.

PR #266, opened 2022-01-04
Fix timestamp comparisons

Olen:
Update will skip every other run, if the previous update took less than a second to complete.
Comparing with ">=" fixes this.

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -893,7 +893,7 @@ class Controller:
                             or vin not in self._last_update_time
                             or (
                                 (cur_time - self._last_update_time[vin])
-                                > _calculate_next_interval(vin)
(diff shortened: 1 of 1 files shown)


Olen:
I see you use time.time() here, which should really not cause any problems.
However, I do see from my debug logs that occationally polling is skipped, and changing this seemingly made that problem go away.

Olen:
I am going to do some more investigations here.  There is something fishy going on somewhere, but I am not sure what happens.
I'll debug the "last_update_time" and a few other variables to find out why it skips some updates.

alandtse:
Ok, I'll mark it as a draft for now. Please swap it back when it's ready for review.

Olen:
Found the root cause of these while adding helper functions.
Fixed in #271

PR #268, opened 2022-01-04
Another ">=" comparison

Olen:
Same issue as #266


alandtse:
Please consolidate with #266

PR #271, opened 2022-01-11
Use helper functions

Olen:
Sorry this looks so intimidating.

This PR incorporates helper functions for setting and getting most private and internal variables. 
I am marking it as a draft, as I am still working on it, but it is getting to a state that actually works pretty well now, so I would appreciate any feedback. 

When I migrated code to the new helper functions, I also needed to update almost all the tests.  That is why so many files have been changed.  All actual code change is in controller.py. The rest are in the different tests.
On the ther hand, almost all tests are now using the real helper functions and code instead of mock functions. This makes the tests more relevant, I hope.

I also updated all the existing helper functions to be consistent in what parameters they accept, and what they return.  And added type hints.

To work on this, I really needed to add and update some more debugging as well, so the logs are now more readable, I hope. 

I also found and fixed a few minor bugs which are strictly not part of the helper functions, but which were fixed for consistency, and because they were revealed when I started using the helper functions.  

Like timestamps sometimes being set as int and sometimes as float.  Now, all timestamps are _rounded_ to int, so the comparisons are more correct.
I found that stuff like below was happening quite frequently when checking if the different intervals between polls had passed:
```
Time: 59.998711585998535 Interval 60, not runnng update
Time: 59.99986433982849 Interval 60, not runnng update
Time: 59.998191595077515 Interval 60, not runnng update
```
This should not be a problem anymore, as we are now always comparing (rounded) ints to ints.
 
Another example of an issue I found while doing this was that sending e.g. a WAKE_UP api call with "wake_if_asleep=True" caused som problems when I started using the helper functions, so that has also been fixed as part of this PR.

All tests and code validation completes without errors locally here, and I am running this tree in my "production" HA.  But I'll keep running it for a few more days while testing different scenarios with the car online, offline, connected, disconnected etc. before I activate the PR.

I'll also remove any lingering, commented out code before I do that.

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -242,8 +242,12 @@ async def wake_up(wrapped, instance, args, kwargs) -> Callable:
         raise
     if valid_result(result):
         _LOGGER.debug("Result: %s", result)
-        _LOGGER.debug("State: %s", result.get("response").get("state"))
-        instance.set_car_online(car_id=car_id, online_status=True)
+        if "state" in result.get("response"):
+            _LOGGER.debug("State: %s", result.get("response").get("state"))
+            instance.set_car_online(
+                car_id=car_id,
+                online_status=result.get("response").get("state") == "online",
+            )
         return result
     raise TeslaException("could_not_wake_buses")
 
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -358,10 +358,8 @@ class Controller:
                 _LOGGER.debug("Skipping car with VIN: %s", vin)
                 continue
 
-            self.__id_vin_map[car["id"]] = vin
-            self.__vin_id_map[vin] = car["id"]
-            self.__vin_vehicle_id_map[vin] = car["vehicle_id"]
-            self.__vehicle_id_vin_map[car["vehicle_id"]] = vin
+            self.set_id_vin(car_id=car["id"], vin=vin)
+            self.set_vehicle_id_vin(vehicle_id=car["vehicle_id"], vin=vin)
             self.__lock[vin] = asyncio.Lock()
             self.__wakeup_conds[vin] = asyncio.Lock()
             self._last_update_time[vin] = 0
@@ -871,7 +869,7 @@ class Controller:
                         asyncio.create_task(
                             self.__connection.websocket_connect(
                                 vin[-5:],
-                                self.__vin_vehicle_id_map[vin],
+                                self.vin_to_vehicle_id(vin=vin),
                                 on_message=self._process_websocket_message,
                                 on_disconnect=self._process_websocket_disconnect,
                             )
@@ -900,10 +898,10 @@ class Controller:
                 cars = await self.get_vehicles()
                 self.car_online = {}
                 for car in cars:
-                    self.__id_vin_map[car["id"]] = car["vin"]
-                    self.__vin_id_map[car["vin"]] = car["id"]
-                    self.__vin_vehicle_id_map[car["vin"]] = car["vehicle_id"]
-                    self.__vehicle_id_vin_map[car["vehicle_id"]] = car["vin"]
+                    self.set_id_vin(car_id=car["id"], vin=car["vin"])
+                    self.set_vehicle_id_vin(
+                        vehicle_id=car["vehicle_id"], vin=car["vin"]
+                    )
                     self.set_car_online(car["vin"], car["state"] == "online")
                     self.car_state[car["vin"]] = car
                 self._last_attempted_update_time = cur_time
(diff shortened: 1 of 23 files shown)


teslajsonpy/controller.py
```diff
@@ -1012,7 +1011,7 @@ def get_climate_params(self, car_id: Text = None, vin: Text = None) -> Dict:
         return self.__climate
 
     def set_climate_params(
-        self, car_id: Text = None, vin: Text = None, params: Dict = {}
+        self, car_id: Text = None, vin: Text = None, params: Dict = None
```
alandtse, line 1014: This is a good fix across the board given the default {} can result in [strange behavior](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python) since it's passing a reference to a dict. Typically you also have to do:
```python
params = params or {}
```
So it'll assign a new dict as the default. However, it might not be necessary if self.__climate[vin] accessors handle the None case.

alandtse (commented):
Thanks for putting this together. Can you confirm changing the default to None doesn't cause issues? It may be as simple as having the internal `params = params or {}` as I discussed.

alandtse:
Looking at the failing test, it's probably ok to ignore the `too-many-statements` for controller.py.

teslajsonpy/controller.py
Olen: I'll look into this.  I don't think there is a problem, but I'll check if we try to iterate the dicts somewhere and maybe just add the extra line as suggested as a security measure.

teslajsonpy/controller.py
```diff
@@ -948,78 +1002,163 @@ async def _get_and_process_energysite_data(energysite_id: Text) -> None:
...
+    def get_charging_params(self, car_id: Text = None, vin: Text = None) -> Dict:
         """Return cached copy of charging_params for car_id."""
-        vin = self._id_to_vin(car_id)
-        if vin:
+        if car_id and not vin:
+            vin = self._id_to_vin(car_id)
+        if vin and vin in self.__charging:
             return self.__charging[vin]
-        return {}
+        return self.__charging
```
alandtse, line 1037: Just wanted to confirm this return is intentional and you've considered the use cases. Previously we'd return an empty dict. now it's returning a dict of dicts when the vin doesn't exist. I'm not opposed if you've thought about it.

teslajsonpy/controller.py
```diff
@@ -948,78 +1002,163 @@ async def _get_and_process_energysite_data(energysite_id: Text) -> None:
...
             self.__charging[vin] = params
 
-    def get_power_params(self, site_id):
+    def charging_state(self, car_id: Text = None, vin: Text = None) -> Text:
+        """Return charging state for a single vehicle."""
+        if car_id and not vin:
+            vin = self._id_to_vin(car_id)
+        if vin and vin in self.__charging:
+            return self.get_charging_params(vin=vin).get("charging_state")
+        return None
```
alandtse, line 1055: Same question about this return. Should it be None or empty dict. Regardless of what we choose, it should be consistent so we don't confuse ourselves later.

teslajsonpy/controller.py
```diff
@@ -1037,12 +1178,15 @@ def get_updates(self, car_id: Text = None):
             processed. Othewise, the entire updates dictionary.
 
         """
-        vin = self._id_to_vin(car_id)
-        if vin:
+        if car_id and not vin:
```
alandtse, line 1181: Should note in docstring that VIN will override car_id.

teslajsonpy/controller.py
Olen: Hi, and thanks for the review.

Yes, this is intentional.  There was some inconsistensies before, where some functions (e.g. `get_updates` and `get_last_update_time` returned the whole dict, while others returned an empty dict, and I wanted to have the same behaviour in all of the functions.
When choosing between different return values, I thought of more (potential) use cases with getting the whole dict, than getting an empty dict.  An other option was to return `None` as that would maybe be the least surprising if no vin or car_id was provided, but again, that would not allow you to get the whole dict, which is useful in cases like `for vin, online in self.get_car_online().items():` (line 960).

teslajsonpy/controller.py
Olen: In this function, I think `None` is the correct return value.  

The functions `is_car_online`, `is_climate_on`, `shift_state`, `charging_state` etc. are shortcuts for getting a single value (or boolean) for one specific car.

I have tried to be consistent, in the sense that the `get_`-functions returns parameters for a single car if vin or car_id is provided and a full dict for all cars if none of these are given.

The `is_` functions are "questions" that can be answered `true` or `false`.

And then the `_state` functions (`shift_state`, and `charging_state`) returns a string with the actual state of a single parameter (given a vin or car_id) or `None`.

(But this made me have another look at the code, and I found that the function "in_gear" should really be "is_in_gear", so I just made another commit to fix that).

I will also update the docstrings to be more clear.

Olen (commented):
I will pull this out again before enabling the PR (unless you think it is ok) as it is not really part of this PR.
I am just testing here, and want to ensure my live testing is in sync with this PR so I don't break anything.

I will also look for a better way to fix the test that failed here.

alandtse:
I'm going to squash it so it's fine.

PR #274, opened 2022-01-16
Always report installed_version in update_sensor

Olen:
Title says it all.

Small patch to lways report the current software version in the update sensor.


alandtse:
Can you please double check we're not using the `None` in the HA component to signify anything? None has a special meaning in HA for "unavailable".

Olen:
I have not found any references to this attribute anywhere else in the code, neither in teslajsonpy nor in tesla_custom.

This just modifies the attributes, not the sensor status. However, I added a check to ensure that we have at least some data from the car, so we don't report attributes for a car that is unavailable.

alandtse:
Ok thanks. You may want to take a look at [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)  so your commit messages don't need to be squashed.

PR #279, opened 2022-01-20
Provide all data in home assistant sensor as json string

ehendrix23:
Not all data available through the API is exposed in the sensors to home assistant. Due to amount of available data it also won't make sense to expose everything as sensors or attributes.
This PR adds a new attribute (vehicle_data) to the binary OnlineSensor containing all the information retrieved from the API as a json string.
This then allows one to extract the data in templates and use as sensors.
For example, to get third row left seat heater in Model X:
```
{{ (state_attr('binary_sensor.<name>_online_sensor', 'vehicle_data') | from_json)['climate_state']['seat_heater_third_row_left'] }}
```

### DIFF
--- a/tests/unit_tests/homeassistant/test_online_sensor.py
+++ b/tests/unit_tests/homeassistant/test_online_sensor.py
@@ -5,9 +5,9 @@ import pytest
 from teslajsonpy.controller import Controller
 from teslajsonpy.homeassistant.binary_sensor import OnlineSensor
 
-from tests.tesla_mock import TeslaMock
+from tests.tesla_mock import TeslaMock, VIN
 
-VIN = "5YJSA11111111111"
+# VIN = "5YJSA11111111111"
 
 
 def test_has_battery(monkeypatch):
@@ -78,14 +78,9 @@ async def test_get_value_on(monkeypatch):
         "id": 12345678901234567,
         "vehicle_id": 1234567890,
         "update_interval": 300,
-        "vehicle_data": {
-            "climate_state": {},
-            "charge_state": {},
-            "vehicle_state": {},
-            "vehicle_config": {},
-            "drive_state": {},
-            "gui_settings": {},
-        }
+        'vehicle_data': '{"climate_state": {}, "charge_state": {}, "vehicle_state": '
+                        '{}, "vehicle_config": {}, "drive_state": {}, "gui_settings": '
+                        '{}}'
     }
 
 
--- a/tests/unit_tests/homeassistant/test_online_sensor.py
+++ b/tests/unit_tests/homeassistant/test_online_sensor.py
@@ -149,5 +149,5 @@ async def test_get_vehicle_data_attribute(monkeypatch):
                                             ' "vehicle_state": {"rt": 0},'\
                                             ' "vehicle_config": {"car_type": "model3"},'\
                                             ' "drive_state": {"shift_state": "P"},'\
-                                            ' "gui_settings": {"gui_range_display": "Rated"}'
+                                            ' "gui_settings": {"gui_range_display": "Rated"}}'
     
--- a/tests/unit_tests/homeassistant/test_online_sensor.py
+++ b/tests/unit_tests/homeassistant/test_online_sensor.py
@@ -5,7 +5,7 @@ import pytest
 from teslajsonpy.controller import Controller
 from teslajsonpy.homeassistant.binary_sensor import OnlineSensor
 
-from tests.tesla_mock import TeslaMock, VIN
+from tests.tesla_mock import TeslaMock, VIN, CAR_ID
 
 # VIN = "5YJSA11111111111"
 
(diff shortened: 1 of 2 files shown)


tests/unit_tests/homeassistant/test_online_sensor.py
```diff
@@ -105,4 +113,7 @@ async def test_get_value_off(monkeypatch):
         "id": 12345678901234567,
         "vehicle_id": 1234567890,
         "update_interval": 300,
-    }
+        "vehicle_data": '{"climate_state": {}, "charge_state": {}, "vehicle_state": '
```
alandtse, line 116: Please have the test mock some test data in addition to the empty data set.

alandtse (changes requested):
Thanks. Let's fill out the tests and then this should be fine. In HA, this may generate a lot of log data so we should check if attributes are saved in the logs. If it is, we may want to flag it as a non default option.

tests/unit_tests/homeassistant/test_online_sensor.py
ehendrix23: Updated test but seem to have issues with mocking the one with data.

ehendrix23:
>  In HA, this may generate a lot of log data so we should check if attributes are saved in the logs. If it is, we may want to flag it as a non default option.

I do not believe attributes are stored in the logs. I cannot see anything in logbook and history does not provide anything for attributes. So I think we're good there.

tests/unit_tests/homeassistant/test_online_sensor.py
alandtse: The tests indicate there is already mock data.

```
E         Full diff:
E           {
E            'id': 12345678901234567,
E            'state': 'asleep',
E            'update_interval': 300,
E         -  'vehicle_data': '{"climate_state": {"climate_state": "on"}, "charge_state": '
E         +  'vehicle_data': '{"climate_state": {"5YJSA11111111111": {"climate_state": '
E         +                  '"on"}}, "charge_state": {"5YJSA11111111111": '
E         -                  '{"charge_state": "on"}, "vehicle_state": {"vehicle_state": '
E         ?                          ^                                 ------------------
E         +                  '{"charging_state": "on"}}, "vehicle_state": '
E         ?                          ^^^              +
E         +                  '{"5YJSA11111111111": {"vehicle_state": "on"}}, '
E         -                  '"on"}, "vehicle_config": {"vehicle_config": "on"}, '
E         ?                   -------                                     -------
E         +                  '"vehicle_config": {"5YJSA11111111111": {"vehicle_config": '
E         ?                                    +++++++++++++++++++++
E         -                  '"drive_state": {"drive_state": "on"}, "gui_settings": '
E         -                  '{"gui_settings": "on"}',
E         +                  '"on"}}, "drive_state": {"5YJSA11111111111": {"drive_state": '
E         +                  '"on"}}, "gui_settings": {"5YJSA11111111111": {"gui_setting": '
E         +                  '"on"}}}',
```
It's likely from this line. https://github.com/zabuldon/teslajsonpy/pull/279/files#diff-033c34e16f4931eef5f977af4319399de9fd75382786738a2c80a2648ea6c619R96

You should be modifying _data like these tests: https://github.com/zabuldon/teslajsonpy/blob/e2dfa6a41f2a1be260bf89065751c1359564792e/tests/unit_tests/homeassistant/test_online_sensor.py#L68

You should have tests that cover:

1. Test for empty data attribute (for example if that values aren't set because an update fails.
2. Test for some actual data.

Ideally these tests should be self contained compared to other tests so if one fails, we can zero in on what behavior is broken. If it's easy to add them as part of an existing test, then that's fine as long as the logic is clear.

Olen:
Any change of the attributes trigger a change in the sensor, so it might fill the history database with updates.  I do not know enough about the details here to be certain how that behaves if just the attributes are changed.

While I can certainly see the use of this, I think it would be even more useful if at least the most relevant information is added as attributes to the already existing sensors and entities. E.g adding `seat_heater_third_row_left` as an attribute to the `sensor.temperature_inside` etc.  
And then maybe move this full dump into another (new) sensor, that one can decide to exclude from HA?

ehendrix23:
> Any change of the attributes trigger a change in the sensor, so it might fill the history database with updates. I do not know enough about the details here to be certain how that behaves if just the attributes are changed.
> 
> While I can certainly see the use of this, I think it would be even more useful if at least the most relevant information is added as attributes to the already existing sensors and entities. E.g adding `seat_heater_third_row_left` as an attribute to the `sensor.temperature_inside` etc. And then maybe move this full dump into another (new) sensor, that one can decide to exclude from HA?

I just checked further on this. Attributes are recorded in the database but only when the state of the entity changes. 
So if an attribute changes but not the state then it is not recorded. If the state itself changes then a new entry is created which includes the attributes.

With the current "proposal" it would thus save the attributes anytime the online sensor changes from on -> off or from off -> on. 
If creating a new sensor then would need to determine what to use for state. Could be something fixed that never changes and then it should never be saved (except maybe when going from unavailable -> available or so) and could be added to be excluded as well.
What should be the state then?

alandtse:
If it's the online sensor, I think the state changes may be infrequent enough that it's ok. My main concern is if the attribute changes itself were triggering log entries. From my perspective, as soon as tests are addressed this can be merged.

To the extent we add any attributes to other sensors, that can wait for later prs. Although, I believe HA is starting to disfavor attribute creation and instead wants sensors.

tests/unit_tests/homeassistant/test_online_sensor.py
ehendrix23: Empty is tested on both test_get_value_on and test_get_value_off. Added a new test test_get_vehicle_data_attribute to specifically test when attributes are filled in.

alandtse:
As a note, I'd recommend you run the test locally in the future so you don't have to keep committing small fixes.

ehendrix23:
> As a note, I'd recommend you run the test locally in the future so you don't have to keep committing small fixes.

Yeah, cause I tested the actual change locally and have it working in my HASS I figured I use the new option of visual studio code to work directly with GitHub without having to clone locally.
It is the tests that's hanging me up. :-) Hopefully this last one works otherwise I'm just going to have to bite the bullet and clone the repository locally and then test locally

alandtse:
The tests aren't necessarily to ensure your code works now. It's to make sure someone in the future doesn't break your code by accident.

alandtse (approved):

PR #281, opened 2022-02-16
Add per vehicle polling interval

ehendrix23:
This PR adds the option for per vehicle polling interval. Through this one can set the polling interval based on each vehicle and thus have a higher polling frequency for 1 vehicle whereas keeping it at a lower polling frequency for another vehicle allowing it to go to sleep.

If no polling interval is set for a VIN then the polling interval is used that has been set for the coordinator.
One can also provide a value of -1 or None as polling interval for a VIN which will then remove that per VIN interval and thus go back to the polling interval for the coordinator.

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -333,6 +333,7 @@ class Controller:
         )
         self.__components = []
         self._update_interval: int = update_interval
+        self._update_interval_vin = {}
         self.__update = {}
         self.__climate = {}
         self.__charging = {}
@@ -772,7 +773,7 @@ class Controller:
                 vin=vin, timestamp=cur_time, shift_state=self.shift_state(vin=vin)
             )
         if self.is_in_gear(vin=vin):
-            driving_interval = min(DRIVING_INTERVAL, self.update_interval)
+            driving_interval = min(DRIVING_INTERVAL, self.get_update_interval_vin(vin=vin))
             if self.__update_state[vin] != "driving":
                 self.__update_state[vin] = "driving"
                 _LOGGER.debug(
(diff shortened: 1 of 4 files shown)


alandtse:
Please add a quick test set for this.

ehendrix23:
> Please add a quick test set for this.

Tests added for polling interval.

alandtse:
Thanks. Given the intent is to allow multiple vehicles to have different intervals, please cover that test case too.

teslajsonpy/controller.py
```diff
@@ -1540,9 +1541,36 @@ def update_interval(self) -> int:
     @update_interval.setter
     def update_interval(self, value: int) -> None:
         """Set update_interval."""
+        if value < 0:
+            value = 300
```
alandtse, line 1545: Sorry, didn't review line by line until now. Create a constant for this and then use it in your test like you have below. We shouldn't have hard coded default numbers buried in the code.

After this, I think it should be good to merge.

ehendrix23: Yeah, I saw it as hard coded hence kept it that way. Changed it to a constant instead. Updated tests as well to test that by default it is set to this constant and for checking.

alandtse: Ooops. I guess we did that a while back if that came from us. Thanks for fixing it. Looks good. I'll merge this an do a release.

ehendrix23: > Ooops. I guess we did that a while back if that came from us. Thanks for fixing it. Looks good. I'll merge this an do a release.

The value = 300 you marked came from me. Did it that way as there was no constant defined for the default in the__init__.
That's what I meant with not hard-coded, did not see it as a constant for the __init__ declaration. :-)

ehendrix23:
FYI, I have changes ready for tesla integration to be able to set polling interval / vehicle. Will open PR once existing PR is done for that.

Issue #295, opened 2022-03-23
Feature request: Homelink support for HA (already WIP in fork)


carleeno:
Hi!
I'm a first time (potential) contributor and wanted to start a discussion and ask some questions before submitting a PR out of the blue.

Now that we have the ButtonEntity in HA, those of us using Tesla's homelink to turn our dumb garage doors into smart garage doors might finally be able to control them in HA.
Related PR #296 
Related tesla_custom integration (draft) PR [here](https://github.com/alandtse/tesla/pull/168)

I have some open questions before I submit a PR though:
- It seems that since opening a garage door is a security item, maybe lock is better than button since extra security measures are built-in to the lock entity. However it'd be pretty cludgy since there's no way to determine or report locked/unlocked and the button will always say "Unlock". This statelessness is why I started with a button entity. Is this ok?
- It gets around the api requirement to have a location value within range of the vehicle by providing the vehicle's own location value. This is the same approach that teslascope uses for it's homelink api call. I think many would value the convenience of the feature over the risk of opening/closing the door when not nearby (plus most if not all powered garage doors have trip and load sensors). However if this would block the feature-add, maybe we can work around this by providing the location from the user's device tracker location from HA via an argument. However this means it can't be remotely triggered, or triggered by a user without a tracker/location. (edit, I'm not sure this workaround is possible as I'm not sure how one would send args with a button entity). Is this ok to allow remotely triggering homelink?
- I'm able to test it with my car which has homelink (and should be done real-world testing later today). I implemented entity availability based on homelink info from the car's state, but I can't test for a vehicle which does not have the homelink device installed at all. Is there someone here who could help test that case?
- Are there any other concerns with adding this entity outside of what I already mentioned?

Thanks!

alandtse:
Hi! Thanks for offering to contribute. It helps since I'm not doing any feature development really.

> Now that we have the ButtonEntity in HA, those of us using Tesla's homelink to turn our dumb garage doors into smart garage doors might finally be able to control them in HA. (edit: still draft) PR #296 Related tesla_custom integration (draft) PR [here](https://github.com/alandtse/tesla/pull/168)
> 
> I have some open questions before I submit a PR though:
> 
> * It seems that since opening a garage door is a security item, maybe lock is better than button since extra security measures are built-in to the lock entity. However it'd be pretty cludgy since there's no way to determine or report locked/unlocked and the button will always say "Unlock". This statelessness is why I started with a button entity. Is this ok?

[...]

PR #296, opened 2022-03-24
feat: add homelink support for HA (#295)

carleeno:
This PR adds a 'trigger homelink" button support for the HA Tesla integration, which is disabled by default.

It works by taking the current location of the vehicle to feed into the required lat/lon params for the API call, which would normally be the user phone's location when called by the official Tesla app.

The button will be disabled by default in HA, and would need to manually be enabled by the the user if they wish to utilize it.

If enabled, it will show as unavailable in these cases:
- if the vehicle does not have the homelink option purchased/installed.
- if the vehicle has no homelink devices programmed.
- if the vehicle is unreachable (as is normal behavior for all entities)

It will, however, show available in all other cases, even if the vehicle is reportedly too far from the homelink device. This is to ensure long polling times wouldn't prevent users from using the button shortly after driving within range of the homelink device.
Pressing the button while the vehicle is actually out of range of the homelink device would result in an error logged in HA (along with any other errors the API request might return).

Tested by:
- creating and running unit tests
- adding to a dev instance of HA with a homelink-equipped Tesla:
  - tested error logging of expected errors (too far from device, etc.)
  - tested real-world ability to open/close a programmed homelink device
  - tested button availability when vehicle is offline, or no homelink devices added.


teslajsonpy/homeassistant/homelink.py
```diff
@@ -0,0 +1,90 @@
...
+        self.hass_type = "button"
+        self.name = self._name()
+        self.uniq_name = self._uniq_name()
+        self.__longitude = None
+        self.__latitude = None
+        self.__homelink_device_count = None
+        self.__homelink_nearby = None
+        self.__homelink_available = False
+
+    async def async_update(self, wake_if_asleep=False, force=False):
```
alandtse, line 43: I'm not sure if HA requires it for buttons but if it doesn't, perhaps we don't need to have an update since we don't have state. It doesn't hurt to have it though I guess.

teslajsonpy/homeassistant/homelink.py
```diff
@@ -0,0 +1,90 @@
...
+            else:
+                self.__longitude = data["longitude"]
+                self.__latitude = data["latitude"]
+        data = self._controller.get_state_params(self._id)
+        if data:
+            self.__homelink_device_count = data["homelink_device_count"]
+            self.__homelink_nearby = data["homelink_nearby"]
+        self.__homelink_available = bool(self.__homelink_device_count)
+
+    def available(self) -> bool:
```
alandtse, line 88: Availability in HA normally means it's network accessible too. Should the test also include a device online test?

teslajsonpy/homeassistant/homelink.py
```diff
@@ -0,0 +1,90 @@
...
+        return False
+
+    async def trigger_homelink(self) -> None:
+        """Trigger Homelink."""
+        if self.__latitude is not None and self.__longitude is not None:
+            if not self.__homelink_device_count:
+                raise HomelinkError(f"No homelink devices added to {self.car_name()}.")
+            if not self.__homelink_nearby:
+                raise HomelinkError(f"No homelink devices near {self.car_name()}.")
+            await self._controller.api(
```
alandtse, line 60: Are there any other homelink specific errors we should consider? For example, what if the lat/long is not near the home? I haven't played with the homelink service so don't know what errors it could produce.

teslajsonpy/homeassistant/homelink.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def refresh(self) -> None:
+        """Refresh data.
+
+        This assumes the controller has already been updated
+        """
+        super().refresh()
+        data = self._controller.get_drive_params(self._id)
+        if data:
+            if data["native_location_supported"]:
```
alandtse, line 76: Don't forget to test both branches.

alandtse (commented):
Just some questions, but overall I think the structure is correct.

teslajsonpy/homeassistant/homelink.py
carleeno: hmm, I'm not sure either. I added it as it was also implemented in alerts.py for the other buttons.
Regardless, wouldn't it be needed to call it's refresh() impl to get homelink availability and location, etc?

teslajsonpy/homeassistant/homelink.py
carleeno: I was looking into this and got stuck, I did discover that when my laptop running the devcontainer went offline, all devices became unavailable except this button, so it should be fixed for sure. However I wasn't able to find out how the other devices determine if they're online.
Any pointers?

teslajsonpy/homeassistant/homelink.py
carleeno: From what I have experienced in my own testing, I've only been able to get a `too_far_from_vehicle` error from the api. I'm sure there's a "no homelink devices added" response but I can't easily test that, I'd have to remove and relearn all of my homelink devices.

Rather than relying on a failed api response error, my approach was to preemptively raise these errors.
Also, since we're pulling lat/lon straight from the car, as long as the car reports a device is nearby, we can be sure we're sending a valid lat/lon.

However I'm not sure if async_update (and thus refresh) run automatically when the button is pressed. If not, there is a possible edge case:
- the latest update reports device nearby and has valid lat/lon; but car has since moved and no longer near a homelink device
  - this would cause the api call to get a response with `{'response': {'reason': 'too_far_from_vehicle', 'result': False}}`, which I'm not sure if the controller would notice this as an error and bubble it up or not. I'd have to manually fudge the code to test this on my car and see what happens. I was under the impression it would get noticed by the controller that the api call returned an error, but now I'm not sure.

teslajsonpy/homeassistant/homelink.py
carleeno: Is it acceptable to reference "private" values in a test? There's no method which returns the lan/lon, so to test this I'd need to assert it's .__latitude and .__longitude are as expected from the test, which seems wrong.

Or is it better to add a method in TriggerHomelink specifically to return these for testing?

carleeno: I partially answered this, with double underscore I can't access them, so I could change to single leading underscore to make them not really private to test, or add a method to return them specifically for testing, whichever you prefer.

alandtse: You can control the value of data. In this case, make the return false.

teslajsonpy/homeassistant/homelink.py
alandtse: I guess since you're storing data, then we do need this. Yes, you'd have to do a refresh to get an instant update. Otherwise there will be a lag between the current state.

teslajsonpy/homeassistant/homelink.py
alandtse: I think the other entities will set availability false if data doesn't return properly. It's been a while so you may just have to look.

teslajsonpy/homeassistant/homelink.py
alandtse: > Rather than relying on a failed api response error, my approach was to preemptively raise these errors.

While it's ok to do this, we should also handle the api errors too even if wrapped up generically. Perhaps see if the result is False and then pass the reason if it exist.

carleeno: I added a response check and raise a HomelinkError in case the response indicates the action wasn't performed to bubble up the error (no matter what it is) to HA logs.

teslajsonpy/homeassistant/homelink.py
carleeno: Thanks, I still wasn't sure of best approach so I took a stab with [a81c574](https://github.com/zabuldon/teslajsonpy/pull/296/commits/a81c574ddee1187fe28e98ae32bb39e4875f331f) by changing the private vars to a single leading underscore so the test can assert them, let me know if that works.

teslajsonpy/homeassistant/homelink.py
carleeno: ahh the device online test is done through the Entity's available property (in the tesla integration, not teslajsonpy), which I initially overrode and forgot to include it's super value in the comparison. 
Fixed in [7b892e2](https://github.com/alandtse/tesla/pull/168/commits/7b892e229017cd4a84955bc0cf647af13fa91437) and tested by forcing the devcontainer offline.

teslajsonpy/homeassistant/homelink.py
carleeno: I added an update (forced) at the start of trigger_homelink to ensure we have fresh location and homelink info before trying to send the api call. this takes care of the edge case mentioned below where you might have just returned home and attempt to trigger homelink before the location is updated.

alandtse: Ok.

teslajsonpy/homeassistant/homelink.py
alandtse: While not necessary since we're at teslajsonpy, I haven't checked, but I assume you are generating a HA log event in the HA component.

teslajsonpy/homeassistant/homelink.py
alandtse: Ok.

teslajsonpy/homeassistant/homelink.py
alandtse: It's better practice not to have to change private variables. You should be able to mock the data info in the test instead. Please reference other tests. (I haven't reviewed the tests here for a while, so if you look and everyone else has just had single underscore variables, then I guess you can follow that trend).

alandtse (commented):
Please double check the mocking of data during tests per my comment. 

I'm also thinking it'd make sense to only generate the sensor if the vehicle has the home_link keys so not to create useless entities. I believe all Teslas have the homelink as an option and no longer standard.

teslajsonpy/homeassistant/homelink.py
carleeno: correct, in HA it's being caught and logged as error.

teslajsonpy/homeassistant/homelink.py
carleeno: Sorry I think we were talking about 2 different things:
I'm not setting private variables, I just needed to read the private variables to assert them in test. Since python 3.10 doesn't allow accessing double underscore private variables I had to rename them to single leading underscore so they can be read in the test.
This is why I was asking is it better to make a new method in TriggerHomelink to return the private values instead of reading them directly from the test. Both options are not great: one is reading private variables, the other is making a method which is never used except in tests, so I wanted to ask which is the least-worst option.

As the test is currently, it's mocking the data, setting the controller to mocked data, updating the button, and ensuring the button's (private) lat/lon values are correct. Is this ok?
```python3
    _data = _mock.data_request_vehicle()
    _data["drive_state"]["native_location_supported"] = 1
    _data["drive_state"]["longitude"] = 12.345
    _data["drive_state"]["native_longitude"] = 23.456
    _data["drive_state"]["latitude"] = 34.567
    _data["drive_state"]["native_latitude"] = 45.678
    _button = TriggerHomelink(_data, _controller)

    _controller.set_drive_params(vin=VIN, params=_data["drive_state"])

    await _button.async_update()

    assert _button._longitude == 23.456
    assert _button._latitude == 45.678
```
(and of course the non-native version was added as well)

carleeno:
> I'm also thinking it'd make sense to only generate the sensor if the vehicle has the home_link keys so not to create useless entities. I believe all Teslas have the homelink as an option and no longer standard.

I initially attempted this, but I discovered that when `_add_car_components` is called to append all the components, we don't yet have good data from the vehicle, meaning we can't determine if it should be added or not.
However I'll investigate making the entity disabled if the keys don't exist.

teslajsonpy/homeassistant/homelink.py
alandtse: Oh sorry.  I guess, the cleaner method would be to have an access function to get all the internal state. While no one is accessing the button state now, it'd be more extensible in the future instead of encouraging protected variable direct access. But honestly, you're right that seems wasteful since it's caused by needing to test. Isn't it just a warning if you try to directly access a private variable during a test? Perhaps we can squelch the warning. I'm ok with what you think makes the most sense in this case. 

I guess the weirdness for me was the fact that this is caching internal state about the vehicle instead of accessing the data directly from the controller when needed.

alandtse:
Ahh, yes, there's some legacy weirdness since the library was created with HA specific items down in the API. #74  If you can easily disable it, then let's go with that, but I won't hold this up just because of that.

teslajsonpy/homeassistant/homelink.py
carleeno: > Oh sorry. I guess, the cleaner method would be to have an access function to get all the internal state. While no one is accessing the button state now, it'd be more extensible in the future instead of encouraging protected variable direct access. But honestly, you're right that seems wasteful since it's caused by needing to test. Isn't it just a warning if you try to directly access a private variable during a test? Perhaps we can squelch the warning. I'm ok with what you think makes the most sense in this case.

Now that it's `_latitude` there's no warning, but originally it was `__latitude` which raised a value error that the variable didn't exist, since it's not accessible at all. 

> 
> I guess the weirdness for me was the fact that this is caching internal state about the vehicle instead of accessing the data directly from the controller when needed.

ahh good point, it doesn't actually need to be cached, but if it's not cached, then I really don't see how to test that it's using the right location in the api call. Apologies for my ignorance as I'm brand new to mocks, but is there a way to intercept the api call to check the parameters? I haven't found any other entity tests here to do that, nor did I find anything like that in TeslaMock.

carleeno:
> Ahh, yes, there's some legacy weirdness since the library was created with HA specific items down in the API. #24 If you can easily disable it, then let's go with that, but I won't hold this up just because of that.

I took a look and I could set disabled by default (like in #299), but it still hits the same problem that this value has to be determined before we have vehicle state, so if we do this, it'd be disabled for all vehicles and wouldn't enable itself once we determine homelink exists.

However maybe it's worth making this always disabled by default anyway, since it seems not to be a common thing, and those who go searching for homelink might find that it exists for them to enable it. It also would alleviate the security concern of adding this button out of the blue to new users.

I'd probably need to wait on #299 to land to prevent merge conflicts if we want to make it disabled by default.

teslajsonpy/homeassistant/homelink.py
alandtse: Take a look at patch. https://docs.python.org/3/library/unittest.mock.html#quick-guide

alandtse: Oh sorry I missed the part about checking parameters. You may be overthinking this. You just need to test that it doesn't throw an exception for being too far when the data includes native locations or not. That will test the two paths. I don't think you have to test the actual internal state.

There may be a way to do what you're proposing with checking the state, but really the goal is to make sure that someone doesn't break your code later and I think black box testing is fine.

carleeno: Unfortunately in unit testing, the lat/lon values have no affect at all, so there's no way to determine if it would throw an exception for being too far, as that's entirely determined by the car itself.

I took a stab at patching controller.api so I could assert_called_with to ensure we're calling with the right lat/lon, however I hit a road block that monkeypatch doesn't offer this feature, and it seems implementing mock/magicmock would break/overwrite our existing TeslaMock/monkeypatch setup.

I also found no other modules in teslajsonpy doing this type of assertion, as they all only inspect what's internal to the lock/switch/climate classes, and don't assert that it's making the right api calls, so you're right that I'm probably overthinking this haha.

if it's ok with you we'll leave it where it's checking the cached lat/lon values are correct, and trust that the api call is made correctly, as the other modules are doing?

alandtse:
I'll wait for you to address the conflict before I push a release.

carleeno:
> I'll wait for you to address the conflict before I push a release.

will take care of that now.
And do we want to go the route of making this button disabled by default for all vehicles?

alandtse:
Disabled by default. We'll want to update the wiki or something on the main component to say to enable it

carleeno:
after merging in dev, make coverage is failing for
```
teslajsonpy/homeassistant/vehicle_data.py:75: in VehicleDataSensor
    exclude_dicts: Optional[list[str]] = None,
E   TypeError: 'type' object is not subscriptable
```
seems due to running python 3.8, this collection type hinting is only supported in 3.9+, I'm not familiar with poetry, but should it define a higher python version requirement?

alandtse:
I think you can import List from typing and avoid that error. But at this point I think we need to support py 3.8.

carleeno:
> I think you can import List from typing and avoid that error. But at this point I think we need to support py 3.8.

True, it can be fixed by using `List` from typing, but this change was from @ehendrix23 (#299 ) which is why I didn't commit the fix in this homelink pr, it should probably be a separate fix PR now that 299 has merged.

alandtse:
Do you want to do the fix and then we can merge this in?

carleeno:
> Do you want to do the fix and then we can merge this in?

sure I'll make a pr quickly

alandtse:
Ok, can you rebase off dev and we can confirm the tests work.

ehendrix23:
Question on this, should 3.8 still be supported? HASS does not support 3.8 anymore with 3.9 being the lowest.
Understand that this API can be used by others as well but just wondering.

alandtse:
Yes.

Issue #298, opened 2022-03-24
Consider moving the `vehicle_data` attribute into a disabled by default entity (`entity_registry_enabled_default=False`)


bdraco:
This sensor attribute which was added in https://github.com/zabuldon/teslajsonpy/pull/279 takes quite a bit of database space because it can't be de-duplicated since at least one of the attributes always changes.  Additionally every client (include mobiles) that connect has to serialize/download/deserialize the full state before the UI loads, and then process every update.

https://github.com/zabuldon/teslajsonpy/blob/dev/teslajsonpy/homeassistant/binary_sensor.py#L220

https://developers.home-assistant.io/docs/core/entity?_highlight=entit#generic-properties
Entities that generate a significant amount of state changes can quickly increase the size of the database when the extra_state_attributes also change frequently. Minimize the number of extra_state_attributes for these entities by removing non-critical attributes or creating additional sensor entities.

https://developers.home-assistant.io/docs/core/entity?_highlight=gener#advanced-properties
`entity_registry_enabled_default=False`

cc @ehendrix23 

As a workaround, I disabled the online binary sensor, but that is one of the ones I did find useful.

ehendrix23:
Hey Nick,

But this is only saved when state changes correct? When I was checking into it there did not seem to be anything saved unless the state itself changed.

Will see to make different sensor(s) (maybe 1 for each area) with disabled by default allowing some more selection then what someone wants enabled.
If indeed only saved on state change, which I assume it is, maybe then further have more of a fixed state to reduce saving and thus database growth?
Thoughts?

bdraco:
Hi Erik,

Thanks for following up 😄 

> But this is only saved when state changes correct? When I was checking into it there did not seem to be anything saved unless the state itself changed.

If any of the attributes change (https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L1353), the `state_changed` event will be fired and the entire set of attributes+state will be written to the database along with the `last_updated` timestamp changes.  The `last_changed` timestamp only changes when the actual state changes (https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L1351).  Both end up firing the `state_changed` event and writing to the db

In 2022.4 and later (not yet released), when the same set of state attributes (exact same) already exists in the database, newly recorded state changes now only store a reference to the state attributes instead of saving the whole set of state attributes.  

[...]

PR #299, opened 2022-03-24
Setup sensors for vehicle data

ehendrix23:
This PR is to create different sensors, disabled by default, to represent all the vehicle data.
Following sensors would be created:

- Climate data sensor
- Charge data sensor
- Vehicle state data sensor
- Software update data sensor
- Speed Limit data sensor
- Vehicle Config data sensor
- Drive State data sensor
- GUI Settings data sensor      

The sensors are disabled by default so that only those whose data is wanted can be enabled. I also separated it into different 
sensors so that is it not an all or nothing.
State for each sensor is set to VIN of vehicle ensuring it is a static value and thus not being saved regularly to the database with all the attributes.
Each element from the response is now it's own attribute and not a JSON string making it easier to retrieve the value.

Tests have been added. This has also been tested within a development HASS instance confirming that:
1. Entities are disabled by default
2. Attributes are provided correctly

I have not removed vehicle_data from online sensor yet. Doing so would be a breaking change and unsure if it should be done now, or first provide this and then later remove it?
 

Fixes Issue #298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,21 +2,11 @@
 
 <!--next-version-placeholder-->
 
-## v1.11.0 (2022-03-25)
+## v1.10.0 (2022-03-25)
 
 ### Breaking
 
 - All vehicle data in home assistant sensor has been removed and new sensors created containing respective data as attributes.
-
-### v1.10.0 (2022-03-25)
-
-### Feature
-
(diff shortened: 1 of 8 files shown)


alandtse:
It'd be a breaking change so let's do it with the same PR. You'd just have to remove the attributes from the online sensor.

ehendrix23:
> It'd be a breaking change so let's do it with the same PR. You'd just have to remove the attributes from the online sensor.

Done. Also have a change for tesla custom component. Will be creating PR shortly.

PR #306, opened 2022-04-24
Remove duplicate ChargeStateDataSensor

ehendrix23:
Sensor ChargeStateDataSensor was added twice to list of components for vehicle resulting in warning within HASS that component is not creating sensors with unique IDs. This did not cause any other issues within HASS.
This PR just removes that duplicate line entry


alandtse:
Oh wow. Thanks

PR #308, opened 2022-04-24
Not all cars have heated steering wheels and seats

megabytemb:
When a car doesn't have a heated steering wheel, teslajsonpy would throw an index error.

simply updating this to a get so it returns None, which is a valid response.

I've also updated heated seats, which used the same code.

Resolves
https://github.com/alandtse/tesla/issues/199


alandtse:
Great thanks!

PR #311, opened 2022-04-26
Disable heated seats and steering wheel by default

ehendrix23:
This PR is to have the heated seats and steering wheel disabled by default.
PR [#205](https://github.com/alandtse/tesla/pull/205) for the tesla component will ensure that those supported by the vehicle will become enabled


alandtse:
Any other PRs likely to come or should this be a release of teslajsonpy?

ehendrix23:
Nothing I'm working on right now. :-)

Pretty sure I'll come up with some more ideas in the future, but I think we're good for this to be released. :-)

PR #314, opened 2022-05-02
Update endpoints.json to app version 4.7.0

Megabytemb:
Hello,

There are many new Commands in the new version of the app, 4.7.0

Just looking to update endpoints.json so we can take advantage of the new commands like activating Dog Mode

new endpoints file gathered from
https://github.com/timdorr/tesla-api/blob/master/docs/miscellaneous/endpoints.md

### DIFF
--- a/teslajsonpy/endpoints.json
+++ b/teslajsonpy/endpoints.json
@@ -1,1688 +1,1688 @@
 {
-	"ACTUATE_TRUNK": {
-		"TYPE": "POST",
-		"URI": "api/1/vehicles/{vehicle_id}/command/actuate_trunk",
-		"AUTH": true
-	},
-	"APP_FEEDBACK_ENTITLEMENTS": {
-		"TYPE": "GET",
-		"URI": "api/1/diagnostics",
-		"AUTH": true
-	},
-	"APP_FEEDBACK_LOGS": {
-		"TYPE": "POST",
-		"URI": "api/1/reports",
-		"AUTH": true
-	},
-	"APP_FEEDBACK_METADATA": {
-		"TYPE": "POST",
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks, did you just do a replace of the endpoints file? I typically just did a [json diff ](http://www.jsondiff.com/)and grabbed new ones. I didn't want to remove endpoint commands unless it's confirmed they're actually gone even if they're deprecated.

Megabytemb:
I simply replaced the file.

Looking at the Diff, there are a few places where they've removed old commands, or updated the URLs, so I'm unsure how you may want to proceed.

If you want to be selective, I'm personally only interested in the following commands being added

- `REMOTE_AUTO_SEAT_CLIMATE_REQUEST`
- `REMOTE_SEAT_COOLING_REQUEST`
- `SET_CLIMATE_KEEPER_MODE`
- `SET_VEHICLE_NAME`

Megabytemb:
After re-thinking about your json-diff comment, I've updated this pull request so it only adds the new commands.

alandtse:
Thanks. Hmm, you saw them change urls for older commands? We should update that. I haven't been removing commands that may be deprecated but in theory could work.

Megabytemb:
They haven't changed any URLS or Methods for commands that we're using. today.

If you're looking for feature parody, there's a little bit to do.
If you're just kind of in maintenance mode, this pull request is adequate.

alandtse:
Ok good enough. We can review any that got changed in the next major update.

alandtse:
I won't take this in the fix release for Tesla, but will add the version bump later. Of course, if you end up needing it, please bump it.

Issue #287, opened 2022-03-13
Unknown Energy Site logic broken since @bassrock 's diff


giachello:
Hi there @bassrock  

your commit broke the logic for handling unknown grid status for me.

 https://github.com/zabuldon/teslajsonpy/commit/5f92fc4e466b78e78cf40bd1fb78d279bba1c09c

I have an active energy site that gives this type of information, Which results in broken reporting (see image below)..

Can you please revert / improve the logic so that it handles my case?

E.g. adding back a condition like:

```
            if "grid_status" in data and data["grid_status"] == "Unknown" and data["solar_power"] == 0:
                return
```

Thank you
Giovanni

```
2022-03-12 17:17:10 DEBUG (MainThread) [teslajsonpy.controller] Updating energysite 184280630899
2022-03-12 17:17:10 DEBUG (MainThread) [teslajsonpy.connection] Token expiration in 32 days, 19:04:42
2022-03-12 17:17:10 DEBUG (MainThread) [teslajsonpy.connection] get: https://owner-api.teslamotors.com/api/1/energy_sites/184280630899/live_status {}
2022-03-12 17:17:10 DEBUG (MainThread) [teslajsonpy.connection] 200: {"response":{"solar_power":332,"timestamp":"2022-03-12T17:17:09-08:00"}}
2022-03-12 17:18:10 DEBUG (MainThread) [teslajsonpy.controller] Updating energysite 184280630899
2022-03-12 17:18:10 DEBUG (MainThread) [teslajsonpy.connection] Token expiration in 32 days, 19:03:42
2022-03-12 17:18:10 DEBUG (MainThread) [teslajsonpy.connection] get: https://owner-api.teslamotors.com/api/1/energy_sites/184280630899/live_status {}
2022-03-12 17:18:10 DEBUG (MainThread) [teslajsonpy.connection] 200: {"response":{"solar_power":0,"grid_status":"Unknown","grid_services_active":false,"timestamp":"2022-03-12T17:18:10-08:00"}}
2022-03-12 17:19:10 DEBUG (MainThread) [teslajsonpy.connection] Token expiration in 32 days, 19:02:42
2022-03-12 17:19:10 DEBUG (MainThread) [teslajsonpy.controller] Updating energysite 184280630899
2022-03-12 17:19:10 DEBUG (MainThread) [teslajsonpy.connection] Token expiration in 32 days, 19:02:42
[...]

PR #317, opened 2022-05-23
Improved handling on 0 Watts spurious power reads

giachello:
Some systems that pre-date Tesla aquisition of SolarCity will have `grid_status: Unknown`, but will have solar power values. At the same time, newer systems will report spurious reads of 0 Watts and grid status "Unknown". In this case, if solar power is 0 return null.

Solves #287

### DIFF
--- a/teslajsonpy/homeassistant/power.py
+++ b/teslajsonpy/homeassistant/power.py
@@ -136,7 +136,11 @@ class PowerSensor(EnergySiteDevice):
         data = self._controller.get_power_params(self._id)
         if data:
             # Note: Some systems that pre-date Tesla aquisition of SolarCity will have `grid_status: Unknown`,
-            # but will have solar power values
+            # but will have solar power values. At the same time, newer systems will report spurious reads of 0 Watts
+            # and grid status unknown. If solar power is 0 return null.
+            if "grid_status" in data and data["grid_status"] == "Unknown" and data["solar_power"] == 0:
+                return
+
             self.__power = data["solar_power"]
             if data["solar_power"] is not None:
                 self.__generating_status = (
--- a/teslajsonpy/homeassistant/power.py
+++ b/teslajsonpy/homeassistant/power.py
@@ -139,6 +139,7 @@ class PowerSensor(EnergySiteDevice):
             # but will have solar power values. At the same time, newer systems will report spurious reads of 0 Watts
             # and grid status unknown. If solar power is 0 return null.
             if "grid_status" in data and data["grid_status"] == "Unknown" and data["solar_power"] == 0:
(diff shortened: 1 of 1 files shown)


alandtse:
Hi @bassrock do you mind checking if this PR does anything that impacts your solar setup? It is closing an issue raised with you original edits. I want to avoid a situation where two people keep reverting each others edits unintentionally.

giachello:
Note that I took @bassrock 's original change and comment into account – it should not affect his case (where status was undefined and watts>0)

bassrock:
@alandtse yea this change is ok 👍

alandtse:
Awesome. Thanks to you both. I'll merge and get a release so it can be updated in the main component.

shred86:
@giachello Sorry, reviving another old commit but this one is causing some issues now that I've implemented getting grid and load power for users that have the Neurio energy monitoring device. I think [this solution](https://github.com/shred86/teslajsonpy/blob/1acc65ce728a1d28e8a1ac4e2f8dcae503b85965/teslajsonpy/controller.py#L1019) will work but wanted to make sure it made sense with you.

Basically I moved the check you added to return `None` to the `get_and_process_site_data` method. The reason being is if we have the case where `grid_status` is `Unknown` and `solar_power` is `0`, I'm simply deleting those two key/values from the response but updating the `self.__power` dictionary with the rest of the data since it's still valid.

I updated the comments to since this is a "normal" condition for my setup which is a Tesla solar system with a Tesla inverter and no Powerwalls. The `grid_status` is always `Unknown` so at night I'm always hitting this case since `solar_power` is `0`. Not a big deal since as soon as we get a `solar_power` value > 0, it will just start updating again.

giachello:
So the challenge is that with my setup, A Delta M with solar panels and no power wall. there are spurious reports during the day, with grid_status Unknown, and power =0  in-between valid readings with power > 0.  If you report that number back instead of null, that will screw up reporting to home assistant. Will your solution work in that scenario?

shred86:
In that situation, it should just continue to use the last valid value so I think it would be the same as it currently is. The `self.__power` dictionary is retained, just the `solar_power` value won't be updated so whatever was reported last, is what HA will continue to see when it refreshes since it's just pulling that dictionary.

shred86:
Also, if you're able to provide the JSON response to `PRODUCT_LIST` and `SITE_DATA`, that would be extremely helpful just to make sure I don't miss anything with your setup. There's some discussion over here about it: https://github.com/zabuldon/teslajsonpy/issues/334

So far we have my setup and a Solar Edge Inverter with a Home Energy Gateway and Powerwall 2. I'm trying to gather as many different setups as I can to see how to best implement stuff for the rewrite.

Issue #320, opened 2022-07-10
error http 406 on every attempted api call


crackers8199:
this error shows up in the log on every single API call i attempt, as of yesterday morning:

```
2022-07-09 18:22:39 DEBUG (MainThread) [teslajsonpy.connection] 406: {"response":null,"error":"not_a_JSON_request","error_description":""}
```

example call, trying to set charging amps to 8:

```
2022-07-09 18:22:39 DEBUG (MainThread) [teslajsonpy.connection] post: https://owner-api.teslamotors.com/api/1/vehicles/xxx/command/set_charging_amps {'charging_amps': 8}
```

Issue #316, opened 2022-05-12
json: cannot unmarshal string into Go struct field OperationUpdateContainer.backup_reserve_percent of type float64 


crackers8199:
trying to run BACKUP_RESERVE on my powerwalls, and getting the following error:

```
2022-05-11 23:15:55 DEBUG (MainThread) [teslajsonpy.connection] post: https://owner-api.teslamotors.com/api/1/energy_sites/xxx/backup {'backup_reserve_percent': '20'}
2022-05-11 23:15:55 DEBUG (MainThread) [teslajsonpy.connection] 400: {"response":null,"error":"https://powergate.prd.sn.tesla.services:443/api/v2/energy_site/backup => Unable to read backup reserve payload --(Error)--> [[ json: cannot unmarshal string into Go struct field OperationUpdateContainer.backup_reserve_percent of type float64 ]]","error_description":""}
2022-05-11 23:15:55 DEBUG (MainThread) [teslajsonpy.controller] Exception: 
api(() {'name': 'BACKUP_RESERVE', 'path_vars': {'site_id': 'xxx'}, 'backup_reserve_percent': '20'})
```

i've tried sending 20, '20', "20", everything i can think of...it always ends up with the same error. it works if i try to run OPERATION_MODE and set it to 'self_consumption' or 'autonomous' so it seems to be an issue with numbers specifically...i tried poking around in the code to see if i could figure it out, but i'm getting nowhere.

i know the actual endpoint does work, because it works if i run it via TeslaPy on the same machine...

```
python3 cli.py -e xxx@xxx.com -a BACKUP_RESERVE -k backup_reserve_percent=20
```

alandtse:
We're literally passing the the kwargs dict straight to httpx and on to tesla's servers. I don't know what calls your cli.py does, but you can compare that with a debugger.

https://github.com/zabuldon/teslajsonpy/blob/dev/teslajsonpy/controller.py#L1801-L1804

It's also quite possible that our endpoints file is also out of date. I don't have a powerwall so can't test anything for you.
https://github.com/zabuldon/teslajsonpy/blob/dev/teslajsonpy/endpoints.json

alandtse:
Btw, the error is saying you can't pass a string, so it has to be `20` not `'20'`

crackers8199:
as I said, I've tried every possible way I can think of (including trying to use a template to cast it). same error every time.

[...]

PR #321, opened 2022-07-10
Switch to json request

haoboji:


haoboji:
resolves alandtse/tesla#231

haoboji:
resolves #320

haoboji:
resolves timdorr/tesla-api#594

alandtse:
Thanks. Did you confirm this doesn't break something like standard login logout?

alandtse:
You should only mark it as closing issues on relevant libraries.  Timdorr's repo is a documentation library so merging this in won't "close" that issue for them.

May also close #316

haoboji:
Hi @alandtse , I've removed the links to other repos. I verified the change in the context of Home Assistant including the situation where access token is refreshed. By looking through the code I did not find anything that may break. It seems to me that the change does not affect login/logout, which invokes `self.websession.get/post` directly instead of the custom ones `self.get/post`.

PR #333, opened 2022-08-03
Initial commit to add grid and load sensors

shred86:
Initial commit to add grid and home (aka load) sensors for Tesla solar systems. All Tesla solar customers with Powerwall should have grid and home reporting as well as non-Powerwall customers with a Tesla inverter and Neurio energy monitoring device, which is what I have. Not all non-Powerwall customers will have a Neurio energy monitoring device as Tesla is not consistent on when they install them. However, I suspect the JSON structure is the same but they will just report back with a value of 0.

Specific changes:
- ~~Added new method `get_site_data` to controller.py. This uses the `SITE_DATA` endpoint where the grid and load power values are located.~~
- ~~During initialization of `Controller`, call `get_site_data` and append it with the `PRODUCT_LIST`~~ Add `load_power` and `grid_power` to `energysite` dictionary so we have valid and correct grid and lower power values when the new `GridPowerSensor` and `LoadPowerSensor` classes are instantiated. Actual values update almost immediately after initialization when `refresh` is called. 
- Created three new classes, `SolarPowerSensor`, `GridPowerSensor` and `LoadPowerSensor`. I kept `PowerSensor` as a parent class to inherit from since the three new power sensors share some attributes.
- Added tests for new code additions.
- Updated some doc strings that I randomly saw were incorrect.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -17,3 +17,4 @@
 -   giachello [Github](https://github.com/giachello)
 -   danielp370 [Github](https://github.com/danielp370)
 -   carleeno [Github](https://github.com/carleeno)
+-   shred [Github](https://github.com/shred86)
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -52,7 +52,11 @@ from teslajsonpy.homeassistant.lock import ChargerLock, Lock
 from teslajsonpy.homeassistant.sentry_mode import SentryModeSwitch
 from teslajsonpy.homeassistant.trunk import FrunkLock, TrunkLock
 from teslajsonpy.homeassistant.heated_steering_wheel import HeatedSteeringWheelSwitch
-from teslajsonpy.homeassistant.power import PowerSensor
+from teslajsonpy.homeassistant.power import (
+    SolarPowerSensor,
+    GridPowerSensor,
+    LoadPowerSensor,
+)
 from teslajsonpy.homeassistant.alerts import Horn, FlashLights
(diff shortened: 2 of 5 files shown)


alandtse:
Thank you. I'm currently on vacation so can't really do a review, but just to confirm for the test cases, you're using values you've basically seen from the API right? I have no solar so cannot verify anything on this side.

shred86:
No worries! Yeah, I essentially just copied them directly from the response. The `ENERGYSITE_CONFIG` mock response is the combination of the `PRODUCTS_LIST` and `SITE_DATA` endpoint, which is one of the changes I made in the `Controller` class because I needed `load_power` and `grid_power`.

alandtse (approved):
Ok, I'm ok with this. Let me know when it's ready to merge (since I noticed you did a push since I last looked).

shred86:
It's good to go now. I removed the `get_site_data` method I added since I realized the existing code is already getting the JSON data from the `SITE_DATA` endpoint almost immediately after initialization. So now I'm just setting default values for `load_power` and `grid_power` to the `energysite` dictionary to initialize, then letting the existing `refresh` method update with the latest values.

One thing to note is if anyone is currently running this, it will be a breaking change in that it will recreate their previous solar generation sensor, likely as a new sensor and they'll just have to delete their one in Home Assistant. Previously it was just a `PowerSensor` class for solar generation, but I created three new classes for the three different power sensor types but left `PowerSensor` as a generic class to inherit from.

alandtse:
Ok thanks. Please note that breaking change in Tesla when you update the dependency to teslajsonpy.

PR #336, opened 2022-08-06
Revert name to prevent breaking change

shred86:
Apologize I didn't catch this earlier but the only reason the previous solar panel sensor was being recreated was because I updated the name to match the naming convention of the other sensors. The name is used as a part of the UUID in Home Assistant which was causing a new solar power sensor to be created. I just reverted back to the original name, solar panel, so it doesn't create a breaking change.

### DIFF
--- a/teslajsonpy/homeassistant/power.py
+++ b/teslajsonpy/homeassistant/power.py
@@ -134,12 +134,12 @@ class SolarPowerSensor(PowerSensor):
     """
 
     def __init__(self, data, controller):
-        """Initialize the solar power sensor."""
+        """Initialize the solar panel sensor."""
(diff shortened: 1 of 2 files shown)


alandtse:
Thanks. If we need to do a breaking, I'm not opposed. Does the old naming convention still make sense even with your changes? We can also revert if you think it doesn't matter.

shred86:
Yeah I think the old name is fine. It's basically "solar panel" instead of "solar power".  I was just trying to align them with what's reported in the JSON response which is `solar_power`, `grid_power` and `load_power` but it's minor. I think it would be better to not introduce a breaking change for something so small.

Issue #334, opened 2022-08-03
"grid_status" and "solar_power" values


shred86:
This is minor but I wanted to document this here to maybe fix in the future. I noticed one of the [comments in power.py](https://github.com/zabuldon/teslajsonpy/blob/8445952897fba6f762d22db384bc1cf68b6b77f3/teslajsonpy/homeassistant/power.py#L142) mentions:
````
            # Note: Some systems that pre-date Tesla aquisition of SolarCity will have `grid_status: Unknown`,
            # but will have solar power values. At the same time, newer systems will report spurious reads of 0 Watts
            # and grid status unknown. If solar power is 0 return null.
````
I have a Tesla solar system with a Tesla inverter (no powerwalls) + Neurio and `grid_status` is always `Unknown`. As such, receiving a `solar_power` value of `0` with `grid_status` of `Unknown` is actually not a spurious energy site power read as currently being logged.

We probably need to get JSON responses from the different Tesla solar configurations and update the [unofficial Tesla API documentation](https://tesla-api.timdorr.com). I've already added the responses for my setup, but there's obviously a wide range of possible setups. These are the various components I'm aware of, and I'm sure I missed a couple. However, I think how the data is reported is largely going to be based on the gateway.

PV arrays:
- Solar panels
- Solar roof

Inverters:
- Tesla inverter (standalone)
- Tesla inverter combined with Powerwall+
- Delta M
- Delta Solivia
- Fronius IG Plus
- Fronius Primo
- Power-One/ABB/Aurora
- SolarEdge
- SolarEdge HD Wave
- SolarEdge HD Wave with SetAPP

Grid monitoring:
- Powerblaster
- Neurio 

Batteries:
- Powerwall
- Powerwall+

Gateways:
- X2e
- Neo
- Series 2
- Tesla inverter (doesn't require a separate gateway)

purcell-lab:
I have SolarEdge with Powerwall, is there a specific script I could run to get the relevant JSON response?

shred86:
> I have SolarEdge with Powerwall, is there a specific script I could run to get the relevant JSON response?

[...]

PR #341, opened 2022-08-18
Update to support Tesla Custom Integration re-write and add battery site support

shred86:
### Overview
This update includes support for the Tesla Custom Integration [re-write](https://github.com/alandtse/tesla/pull/250) and adds support for Tesla solar systems with Powerwall. The general design I'm following is what I see from other libraries like [simplisafe-python](https://github.com/bachya/simplisafe-python) and Home Assistant [documentation](https://developers.home-assistant.io/docs/api_lib_index/). I'm not a developer by any means, so please provide feedback/recommendations. The general flow for using teslajsonpy is:

1. Create an instance of `teslajsonpy.Controller`.
2. Call `Controller.connect` to get a list of products from Tesla to create `self._vehicle_list` and `self._energysite_list`. Two new arguments available, `include_vehicles` and `include_energysites` which defaults True.
3. Call `Controller.generate_car_objects` which returns a dictionary of car objects by VIN number.
4. Call `Controller.generate_energysite_objects` which returns a dictionary of energysite objects by energy site ID.

The car and energy site objects are the primary means to get car and energy site data and send commands.

### Changes
- New module `car.py` that contains a `TeslaCar` class.
- New module `energy.py` that contains energy site classes.
- New method `Controller.generate_car_objects` that generates `TeslaCar` objects and stores them into `self.cars` by vin.
- New method `Controller.generate_energysite_objects` that generates `SolarSite`, `PowerwallSite` or `SolarPowerwallSite` objects and stores them in the `self.energysites` dictionary by `energysite_id`.
- Modified `Controller.update` to now just send a single request to the `PRODUCT_LIST` endpoint (instead of both `PRODUCT_LIST` and `VEHICLE_LIST` to get all products on a Tesla account. From there, create a list of cars `self._vehicle_list` and energy sites `self._energysite_list`.
- Storing JSON responses as-is into their own dictionary by VIN for cars and energysite_id for energysites. These are then passed to `TeslaCar` and `EnergySite` when instantiated.
- Added `include_vehicles` and `include_energysites` arguments to `Controller.connect` which default to True. This provides the option to completely ignore vehicles or energysites.
- Removed `Controller.command`, `Controller.get` and `Controller.post` methods. Only `Controller._wake_up` was using `post` which now uses `Controller.api`.
- Removed all Home Assistant specific modules.
- Removed some unused code specific to energy sites.
- Changes to some naming to try to better align with what the Tesla API uses.

### Fixes
- #348
- #334 
- #24 

### Energy sites
For energy sites, there's a few variations that a user could have and we still need to get more JSON response examples from as many different setups as possible. My guess is the gateway is what determines how the data is published, whether it's a Tesla inverter, home energy gateway, Powerwall 2, etc. What I can tell right now is:

- Solar panels with a Tesla inverter (no Powerwalls), the data we need is from the `SITE_DATA` endpoint.
- Solar panels with a Solar Edge inverter and Home Energy Gateway with Powerwall 2, the data we need comes from `BATTERY_DATA` and `BATTERY_SUMMARY`

There's a couple variations on solar setups but in general, I'm treating it as essentially three from a Class design perspective:
- Solar only
- Powerwall only
- Solar and Powerwall

As you can see in the changes, I created a base `EnergySite` class and three child classes,`SolarSite`, `PowerwallSite` and `SolarPowerwallSite`. The first two inherit from `EnergySite` and `SolarPowerwallSite` inherit from `SolarSite` and `PowerwallSite`. There's probably a better way to do this, but this is what I came up with so far.

### Notes
- Currently passing in the entire instance of the `Controller` (self) when calling `Controller._generate_car_objects` so we have access to all methods on the Controller compared to energysites, I'm only passing in `self.api`, the specific energysite in `self.energysites` and `self.__power_data[energysite_id]`.
- Trunk, frunk and charger doors aren't really "locks" so I find it a bit confusing that if I'm "unlocking" a trunk, it opens. I'm sure this was done intentionally but I wonder if something like a Cover entity would be better.

### To Do / Request Help
- Does the username and password authentication actually work? It would be nice to be able to allow users to authenticate with username and password instead of passing a refresh token in HA.
- Testing methods is simply asserting `is None` since they don't return anything. I'm assuming we probably want to do something like check if `Controller.api` is being called, but `unittest.mock.patch` doesn't work with async functions.
- Need to update the sphinx-docs. I updated some of the `.rst` in `/docs/teslajsonpy` based on my understanding of skimming over the sphinx documentation but I haven't figured out how to build and update everything.


teslajsonpy/controller.py
```diff
@@ -390,6 +395,7 @@ async def connect(
         wake_if_asleep: bool = False,
```
shred86, line 395: Is `wake_if_asleep` even used? When looking at the `Controller.api`, `Controller.post` and `Controller.get` methods, I see it's parameter but not used anywhere.

alandtse: It's used by a wakeup decorator. Search and you should see it.

shred86: Ah got it. I'll have to spend some time understanding how that works!

shred86:
@alandtse Should I go ahead and remove any code that won't be used with the Tesla Custom Integration rewrite? I don't foresee this version of teslajsonpy being used with the pre-rewrite version of Tesla Custom Integration so my thought is to just remove all the HA specific stuff.

alandtse:
> Should I go ahead and remove any code that won't be used with the Tesla Custom Integration rewrite? I don't foresee this version of teslajsonpy being used with the pre-rewrite version of Tesla Custom Integration so my thought is to just remove all the HA specific stuff.

Yes. The intent is the HA stuff needs to live in the tesla component. Tests should also migrate over. That will allow us to make teslajsonpy an async library that others can really use.

shred86:
The only place I see `Controller.post` being used is with `Controller._wake_up`. However, I don't see any reason why `Controller.api` can't be used here to send the `WAKE_UP` post request like I'm doing in `TeslaCar`. I don't see `Controller.get` or `Controller.command` being used anywhere as everything appears to be using `Controller.api`. I made the change in `Controller._wake_up` to use `api` instead of `post` and everything appears to work the same as before. I also removed the `post`, `get` and `command` methods. Is there anything I'm missing as to why we need to keep these three methods?

alandtse:
API was added much later. It could be a replacement if someone doesn't need the low level control from before. Wakeup was written before API.

teslajsonpy/controller.py
```diff
@@ -406,61 +379,81 @@ async def connect(
...
+
+                if energysite[RESOURCE_TYPE] == RESOURCE_TYPE_BATTERY:
+                    self._battery_data[energysite_id] = {}
+                    self._battery_summary[energysite_id] = {}
+                # For dealing with sites that always report "Unknown"
+                # Default to True and check during updates
+                self._grid_status_unknown = {energysite_id: True}
+
+                self.__lock[energysite_id] = asyncio.Lock()
+                self.__lock[battery_id] = asyncio.Lock()
```
shred86, line 449: Do we actually need a separate `asyncio.Lock()` based on `battery_id`? I have no experience using `asyncio.Lock()` but just based on the readings, I'm thinking it's not necessary. The only place I use `self.__lock[battery_id]` is with `Controller.update._get_and_process_battery_data` and ``Controller.update._get_and_process_battery_summary` methods, but only because I was basically copying the structure of previous methods.

I think with those two methods, I can just use the `self.__lock` by `energysite_id` that we already have.

alandtse: The lock is for protecting a resource from access if multiple callers may be reading and writing it. Is there a situation where the library may be doing that for energy sites? Like would multiple entities be potentially making calls to teslajsonpy at the same time?

shred86: I think it's rare but possible. Some folks do have two Tesla solar systems for a single home. This usually happens either if the customer has a very big setup or if they want additional panels after they already had a system installed, they'll just install a separate system so within the Tesla app, you'll see two separate systems.

In this situation, I think it would essentially be like having two cars when it comes to making calls to teslajsonpy. If that's the case, it seems like we could just use `self.__lock[energysite_id]` for all energy site calls but that's what I'm not entirely sure on (if we actually need a separate lock by `battery_id`).

alandtse: I don't know why it was added. For cars it's because all cars are updated every time we hit the end point. I wanted to avoid multiple cars triggering multiple overlapping updates. Maybe locks aren't needed anymore.

shred86: Ah okay, that makes sense for cars. I'm actually not sure what the behavior is if someone has two systems. I just assumed each system would have its own `energysite_id`, and each time we hit the end point we're specifying the `energysite_id`. I think I'm just going to remove them for now since 1) it's a pretty rare case for solar setups and 2) based on the current (assumed) understanding, we would have to hit each endpoint anyways to get the data for each system. Appreciate the info!

teslajsonpy/controller.py
```diff
@@ -1146,11 +1045,6 @@ def charging_state(self, car_id: Text = None, vin: Text = None) -> Text:
             return self.get_charging_params(vin=vin).get("charging_state")
         return None
 
-    def get_power_params(self, site_id: Text) -> Dict:
-        """Return cached copy of power_params for site_id."""
-        energysite_id = self._id_to_energysiteid(site_id)
-        return self.__power[energysite_id]
-
     def get_state_params(self, car_id: Text = None, vin: Text = None) -> Dict:
```
shred86, line 1048: What was the reasoning for being able to get car parameters using `car_id`? With how I'm restructuring this rewrite, I'm just storing the entire `VEHICLE_DATA` response to `self._vehicle_data` by VIN and passing that to the new `TeslaCar` class when it's instantiated. That makes most of these `get_*` and `set_*` methods not needed anymore but I just want to make sure I'm understanding what the purpose was for being able to use `car_id` or `vin` to get and set car data.

alandtse: We allowed getting a single cars info instead of multiple.

shred86: You can do that with the VIN though and everywhere I see these methods being called, it's used with the vin argument.

alandtse: You'd have to check the history. The vin may not have been an option when we wrote it.

shred86:
I just noticed the endpoint `CACHED_PROTO_VEHICLE_DATA` which has a note "This is cached data, pushed by the vehicle on sleep, wake and around OTAs.". When looking at the JSON response of this endpoint, it seems to have everything `VEHICLE_DATA` has but even more. I also noticed everything in the `VEHICLE_DATA` response is also in the `CACHED_PROTO_VEHICLE_DATA` response but listed under a `legacy` key, which makes me think perhaps using `CACHED_PROTO_VEHICLE_DATA` might be better.

Anyone happen to know the difference between these two? If the note is accurate, it seems like to me we could simply just poll this endpoint without having to worry about waking up the vehicle. If it's asleep, it _should_ just return the cached data.

alandtse:
That seems new. We could use that to start but do we lose the info on whether a car is sleeping? I guess the online sensor may be enough to figure that out.

This may also solve the issue of https://github.com/alandtse/tesla/issues/251.

I know you've been basically working yourself on this for like a month without my help. Are we getting to a point where we should consider taking it live?

shred86:
It appears to have the car state as well `"state": "asleep"`, but after looking at this a bit more it appears everything listed under the `data` key might only be updated on certain events (i.e. on sleep, on wake, etc.) as described in the notes. As an example, I started charging my car and waited about 5 minutes. All of the information under `data` appears to be old (charging_state showing complete) but everything under `legacy` is current (charging_state showing charging). This might still be useful to just use this endpoint and if the car is awake, pull the data from the `legacy` key and if it's asleep, use the `data` key.

I've been running it 24/7 on a separate HA install and it's been working well so far. I think it's ready for use. If we could also release a beta version of the HA Tesla integration, that would be good to get some more folks to maybe test it out. There's a couple things I'd like to clean up on the integration though first but I'll continue to update that PR.

alandtse:
Once it's in Dev it's basically beta, but you'll need to recruit people willing to test it.

> This might still be useful to just use this endpoint and if the car is awake, pull the data from the `legacy` key and if it's asleep, use the `data` key.

This sounds fine. It can be the backup source when sleeping.

shred86:
Ah bummer, it looks like `CACHED_PROTO_VEHICLE_DATA` returns 408 (vehicle unavailable) as well when the car is asleep, so that's not going to work.

alandtse:
Just so it's tracked, please ensure conflicts are resolved so it merges cleanly. I can review and provide comments then. This should also make sure we don't obliterate any fixes that have been submitted by others. You can just look at closed PRs as there haven't been too many.

teslajsonpy/__init__.py
```diff
@@ -5,61 +5,30 @@
 For more details about this api, please refer to the documentation at
```
alandtse, line 5: Please make sure we add the header including the SPDX-License-Identifier to every file.

docs/teslajsonpy/teslajsonpy.energy.rst
```diff
@@ -0,0 +1,10 @@
+==========================
```
alandtse, line 1: Is the make docs broken? This seems really light. The CI for making docs doesn't really work so we shouldn't rely on it especially given the huge changes you have. I believe your functions/classes look documented enough.

alandtse (changes requested):
I assume the tests we had for the HA components will be migrated to the /tesla app right? After you add the headers in and confirm the documentation state, I'll do a breaking change release based on your first post. We can then integrate your PR into `dev` and let it run for a week or two. We'll want to do the official release to master when you are available to handle any major bugs so let me know what your preferred schedule is. I'm based in California and we can target a weekend.

Thanks again for all your hard work on this. I can't adjust your status on this repo, but I can make you a maintainer on tesla if you'd like to help out more actively.

teslajsonpy/__init__.py
shred86: Done.

docs/teslajsonpy/teslajsonpy.energy.rst
shred86: Docs are one area I'm not really sure how to update. I noticed we're using sphinx docs and I tried looking through the documentation to figure it out but didn't get very far and could use help with.

shred86:
> I assume the tests we had for the HA components will be migrated to the /tesla app right? After you add the headers in and confirm the documentation state, I'll do a breaking change release based on your first post. We can then integrate your PR into `dev` and let it run for a week or two. We'll want to do the official release to master when you are available to handle any major bugs so let me know what your preferred schedule is. I'm based in California and we can target a weekend.
> 
> Thanks again for all your hard work on this. I can't adjust your status on this repo, but I can make you a maintainer on tesla if you'd like to help out more actively.

That sounds good. I'm in the same time zone so that should work. I'm happy to continue helping out when I can so more than happy to be a maintainer on the Tesla integration.

docs/teslajsonpy/teslajsonpy.energy.rst
alandtse: I fixed an indent and ran `make docs`. I've pushed it to your repo so this should be resolved. Please do fix the failing tests though. `make test`

shred86: Thanks. I'm showing all the tests are passing with `make test`.

alandtse: Odd, my local is still reporting a test failure but the CI online seems to be ok. I'll merge this in and see if that PR triggers it. Assuming it doesn't, this will be the breaking change for us in teslajsonpy.

PR #346, opened 2022-08-28
Fix for grid_status case

shred86:
Fixes an issue that was identified where `grid_status` is actually not reported when the state is `Active` in certain setups.

Sorry for all the changes @alandtse. Not having JSON responses on the various energy site setups and cases is proving to be a pain.

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -988,7 +988,10 @@ class Controller:
                     data = None
                 if data and data["response"]:
                     response = data["response"]
-                    if response["grid_status"] == "Active":
+                    if (
+                        "grid_status" not in response
+                        or response.get("grid_status") != "Unknown"
+                    ):
                         self._grid_status[energysite_id]["grid_always_unk"] = False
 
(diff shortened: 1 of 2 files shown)


teslajsonpy/controller.py
```diff
@@ -988,7 +988,10 @@ async def _get_and_process_energysite_data(energysite_id: Text) -> None:
                     data = None
                 if data and data["response"]:
                     response = data["response"]
-                    if response["grid_status"] == "Active":
+                    if (
+                        "grid_status" not in response
```
alandtse, line 992: FYI, another way to write this is to do the get to a variable that you then test. You test it's not None and then test it's value. For example, in the other part of this change you have already gotten 'solar_power' as __solar_power. Just run your test on that to avoid redundant gets.

shred86: That makes sense but in this situation, should I just use get every time because the response does not have the `grid_status` key if the status is actually `Active` but it does have the key when it's `Unknown`, so it varies even with the same system when the key exists.

PR #349, opened 2022-09-19
switch trunk, frunk, battery door lock types to cover type

### LINKED ISSUE #348: Change Trunk, Frunk and Battery Door to be cover and not lock
I think these entities make more sense as the type `cover` with a class of `DEVICE_CLASS_DOOR` vs. locks as it aligns to how they work better (these controls open and close something and have a status).

chrisdrackett:
fixes #348 

I'm somewhat new to this, and don't know how to test, but I did my best to at least get this started.

### DIFF
--- a/teslajsonpy/homeassistant/lock.py
+++ b/teslajsonpy/homeassistant/lock.py
@@ -8,7 +8,7 @@ https://github.com/zabuldon/teslajsonpy
 import time
 
 from teslajsonpy.homeassistant.vehicle import VehicleDevice
-
+from homeassistant.components.cover import CoverEntityFeature
 
 class Lock(VehicleDevice):
     """Home-assistant lock class for Tesla vehicles.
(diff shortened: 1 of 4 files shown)


teslajsonpy/homeassistant/lock.py
```diff
@@ -118,13 +118,15 @@ def __init__(self, data, controller):
...
-        self.type = "charger door lock"
-        self.hass_type = "lock"
+        self.type = "charger door"
+        self.hass_type = "cover"
 
         self.name = self._name()
 
         self.uniq_name = self._uniq_name()
         self.bin_type = 0x7
+        self.device_class: Text = "DEVICE_CLASS_DOOR"
```
alandtse, line 128: Is Text imported? It's causing a linter error.

alandtse (commented):
Thanks for this. This is a breaking change so I'm reluctant to bring it in while we have another major breaking change being developed. @shred86 how's your ETA for your changes? Or you could just merge this PR into your set of changes.

shred86:
> Thanks for this. This is a breaking change so I'm reluctant to bring it in while we have another major breaking change being developed. @shred86 how's your ETA for your changes? Or you could just merge this PR into your set of changes.

The rewrite does have these entities as cover types. I think I've done about as much as I can with the rewrite at this point. I've been using it on a separate HA instance for several weeks and it seems to be working well.

alandtse:
This should be handled in by #341

PR #351, opened 2022-10-07
Add fix for api change in version 2022.36 (missing charge_to_max_range)

mariogrip:
Version 2022.36 seems to not include charge_to_max_range.

This uses charge_limit_soc[_max] to determine it if charge_to_max_range is not available

### DIFF
--- a/teslajsonpy/homeassistant/charger.py
+++ b/teslajsonpy/homeassistant/charger.py
@@ -118,11 +118,9 @@ class RangeSwitch(VehicleDevice):
         if last_update >= self.__manual_update_time:
             data = self._controller.get_charging_params(self._id)
             if data:
-                """
-                Version 2022.36 seems to not include charge_to_max_range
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks!

alandtse:
Thanks. There's a lint error due to a use of a string that isn't a comment. Please make it a comment and not a multiline string.

alandtse:
Thanks. I consider this ready to merge. Is this backward compatible with people on the older api you think? I haven't run into the keyerror yet.

mariogrip:
> Thanks. I consider this ready to merge. Is this backward compatible with people on the older api you think? I haven't run into the keyerror yet.

Yes it is, it will try to use the older charge_to_max_range first, if is not there then it will use the newer charge_limit_soc[_max]

alandtse:
Thanks. We'll also need to update the tesla to use the new version that I'll generate in a second.

PR #355, opened 2022-10-22
fix: check energysite_id in _grid_status_unknown

alandtse:


shred86:
I'm concerned this might just be masking the issue that was reported. When we call `generate_energysite_objects` in the HA integration, it's adding the `energysite_id` to the `self._grid_status_unknown` dictionary. This occurs before anything else happens so unless `energysite_id` doesn't exist in the `PRODUCT_LIST` endpoint, I don't see how we could have a key error here. If that `energysite_id` key is missing, nothing else is going to work. I'll keep an eye on this one but we may need @bassrock to post the `PRODUCT_LIST` endpoint response to confirm.

@bassrock I'm assuming you'll still run into more issues even after the Tesla Custom Integration is bumped to use this teslajsonpy version. Please post any errors you get in a new issue in this repo and I'll try to help from there.

alandtse:
Even if masking, a KeyError is an integration breaking error. I'd prefer to have a subtle error that doesn't break the integration then rely on having a breaking error to know something is wrong. I can't maintain energysites but at least can help with errors like this.

PR #356, opened 2022-10-26
feat: add tpms, door and window properties

shred86:
Adds tire pressure monitor sensors, doors and window properties.


alandtse:
I'm ok with this in general, but wonder if adding entities to HA should require making edits to teslajsonpy. The point of the rework was to not require stuff like this anymore.

PR #357, opened 2022-10-28
feat: Add window control and window binary sensor

InTheDaylight14:
This PR adds a is_window_open parameter that will return 1 if any window is open. It also adds a close_windows and vent_windows function. This has been working from Home Assistant with the custom_api calls. 

I need to figure out how to run the tests. ~~"make init" in terminal did not recognize "make". Any guidance on the tests or the code is appreciated.~~

~~Got make installed fine, receiving this error "Expecting value: line 1 column 1 (char 0)". I'll keep poking.~~

All good now, tests are working on my end.

### DIFF
--- a/teslajsonpy/car.py
+++ b/teslajsonpy/car.py
@@ -546,8 +546,8 @@ class TeslaCar:
             or self._vehicle_data.get("vehicle_state", {}).get("rd_window")
             or self._vehicle_data.get("vehicle_state", {}).get("rp_window")
         ):
-            return 1
-        return 0
+            return True
+        return False
 
     async def _send_command(
         self, name: str, *, path_vars: dict, wake_if_asleep: bool = False, **kwargs
@@ -988,8 +988,8 @@ class TeslaCar:
             "WINDOW_CONTROL",
             path_vars={"vehicle_id": self.id},
             command="close",
-            lat=0,
-            long=0,
+            lat=self.latitude,
+            long=self.longitude,
             wake_if_asleep=True,
         )
         if data and data["response"]["result"] is True:
--- a/teslajsonpy/car.py
+++ b/teslajsonpy/car.py
@@ -547,8 +547,7 @@ class TeslaCar:
             or self._vehicle_data.get("vehicle_state", {}).get("rp_window")
         ):
             return 1
-        else:
-            return 0
+        return 0
 
     async def _send_command(
         self, name: str, *, path_vars: dict, wake_if_asleep: bool = False, **kwargs
@@ -965,7 +964,7 @@ class TeslaCar:
             self._vehicle_data["vehicle_state"].update(params)
 
     async def vent_windows(self) -> None:
-        """Vent Windows"""
+        """Vent Windows."""
         data = await self._send_command(
             "WINDOW_CONTROL",
             path_vars={"vehicle_id": self.id},
@@ -984,7 +983,7 @@ class TeslaCar:
             self._vehicle_data["vehicle_state"].update(params)
 
     async def close_windows(self) -> None:
-        """Close Windows"""
+        """Close Windows."""
         data = await self._send_command(
             "WINDOW_CONTROL",
             path_vars={"vehicle_id": self.id},
--- a/teslajsonpy/car.py
+++ b/teslajsonpy/car.py
@@ -537,6 +537,19 @@ class TeslaCar:
         """Return rear passenger window status."""
         return self._vehicle_data.get("vehicle_state", {}).get("rp_window")
 
+    @property
+    def is_window_open(self) -> bool:
+        """Return car is window open."""
+        if (
+            self._vehicle_data.get("vehicle_state", {}).get("fd_window")
+            or self._vehicle_data.get("vehicle_state", {}).get("fp_window")
+            or self._vehicle_data.get("vehicle_state", {}).get("rd_window")
+            or self._vehicle_data.get("vehicle_state", {}).get("rp_window")
+        ):
+            return 1
+        else:
+            return 0
+
     async def _send_command(
         self, name: str, *, path_vars: dict, wake_if_asleep: bool = False, **kwargs
     ) -> dict:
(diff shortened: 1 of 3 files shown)


alandtse (changes requested):
Thanks. @shred86 I want your thoughts too about doing these properties in teslajsonpy. We wanted to avoid making people edit the API library directly to edit HA. We should nip it in the bud if we don't intend to have these double commit requirements at both the API and HA level.

teslajsonpy/car.py
```diff
@@ -950,3 +962,41 @@ async def unlock(self) -> None:
...
             params = {"locked": False}
             self._vehicle_data["vehicle_state"].update(params)
+
+    async def vent_windows(self) -> None:
+        """Vent Windows."""
+        data = await self._send_command(
+            "WINDOW_CONTROL",
+            path_vars={"vehicle_id": self.id},
+            command="vent",
+            lat=0,
```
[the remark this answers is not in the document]
InTheDaylight14: After reviewing the [parameter docs](https://tesla-api.timdorr.com/vehicle/commands/windows) again looks like 0 is okay for venting but they should be close to current location for closing. I missed that.

shred86:
I think the only other way to do it is expose the entire vehicle_data JSON response to HA, and pull from the keys there. That would require a bit of work to redo. However, I liked the idea of doing properties for each attribute on the teslajsonpy side so we can modify the output to the integration if needed. For example, the battery state for energy sites has an issue where the Tesla API will randomly and often report 0 for one update. This is something where I like that we can “fix” it in one place (assuming multiple projects were using this library).

I fully expected we would continue to add to teslajsonpy as we went along, especially in the early stage. I didn’t expose everything with rewrite but I figure over time as we keep adding to this project, we’ll get to a point where we’ve added about everything you could possibly want. At that point, it would just be small changes to react to changes with the Tesla API.

teslajsonpy/car.py
```diff
@@ -537,6 +537,18 @@ def window_rp(self) -> int:
         """Return rear passenger window status."""
         return self._vehicle_data.get("vehicle_state", {}).get("rp_window")
 
+    @property
+    def is_window_open(self) -> bool:
```
shred86, line 541: This is minor but the other properties are checking if things are closed, `is_frunk_closed`, `is_trunk_closed`, etc. Just to match everything else, maybe switch this to `is_window_closed` and update the returns to reflect the state.

teslajsonpy/car.py
```diff
@@ -950,3 +962,41 @@ async def unlock(self) -> None:
...
+            "WINDOW_CONTROL",
+            path_vars={"vehicle_id": self.id},
+            command="vent",
+            lat=0,
+            long=0,
+            wake_if_asleep=True,
+        )
+        if data and data["response"]["result"] is True:
+            params = {
+                "fd_window": 1,
```
shred86, line 978: I was actually curious about this - does the Tesla API JSON response actually return 1? Or is it 255 for a fully open window and anything between 0-255 for a partially open window? That seems to be the case with the trunk.

alandtse:
> I think the only other way to do it is expose the entire vehicle_data JSON response to HA, and pull from the keys there. That would require a bit of work to redo. However, I liked the idea of doing properties for each attribute on the teslajsonpy side so we can modify the output to the integration if needed. For example, the battery state for energy sites has an issue where the Tesla API will randomly and often report 0 for one update. This is something where I like that we can “fix” it in one place (assuming multiple projects were using this library).

I thought that was the intent of #24 . That the raw json would be accessible to the software using teslajsonpy so they can construct their own entities. Any smoothing would be the responsibility of the software using the library. Adding too much complexity here suddenly makes it harder for consumers to build on top of this because the paradigm may be different.  If we're back to the, every consuming library must edit the underlying API, then we may not have gained anything from the breaking change. The main issue we had is no one wanted to contribute to Tesla because they'd have to figure out two separately libraries and commit changes in each.

That said, if we're committed to having everyone add properties here, then I guess that's fine since you're basically maintaining it now and the fact it's on a feat PR may mean it's easy enough for new contributors. @InTheDaylight14 I'd be interested in your thoughts on how easy it is to modify the HA component (assuming that's why you're contributing).

shred86:
Agreed with the intent and that's what I was thinking as well. I guess my point is this rewrite is not finished. The end state is everything in the Tesla API is exposed as a property so other software (e.g. HA) can integrate whatever they see fit. I just didn't have the time to do everything and only implemented what was needed for the Tesla integration rewrite (plus a couple extras) for the time being.

alandtse:
> The end state is everything in the Tesla API is exposed as a property so other software (e.g. HA) can integrate whatever they see fit.

That is a lot of work and until you've mapped it all out, will result in the double commit issue I was referencing. Take a look at what I'm saying in the architecture thread. We can consider more there. https://github.com/alandtse/tesla/discussions/316

InTheDaylight14:
> > I think the only other way to do it is expose the entire vehicle_data JSON response to HA, and pull from the keys there. That would require a bit of work to redo. However, I liked the idea of doing properties for each attribute on the teslajsonpy side so we can modify the output to the integration if needed. For example, the battery state for energy sites has an issue where the Tesla API will randomly and often report 0 for one update. This is something where I like that we can “fix” it in one place (assuming multiple projects were using this library).
> 
> I thought that was the intent of #24 . That the raw json would be accessible to the software using teslajsonpy so they can construct their own entities. Any smoothing would be the responsibility of the software using the library. Adding too much complexity here suddenly makes it harder for consumers to build on top of this because the paradigm may be different. If we're back to the, every consuming library must edit the underlying API, then we may not have gained anything from the breaking change. The main issue we had is no one wanted to contribute to Tesla because they'd have to figure out two separately libraries and commit changes in each.
> 
> That said, if we're committed to having everyone add properties here, then I guess that's fine since you're basically maintaining it now and the fact it's on a feat PR may mean it's easy enough for new contributors. @InTheDaylight14 I'd be interested in your thoughts on how easy it is to modify the HA component (assuming that's why you're contributing).

Yeah, I had looked at making the window contribution to the HA component a while ago. I was tuned away pretty quickly not understanding how the two libraries interfaced, and not knowing how to setup a new entity type. A big reason I can now make this contribution is that other covers have been added and I can use them as an example for a new entity. But I also think that it is generally easier to understand now. I'm only a hobby programmer so having tweaks be possible for me is a good sign.

PR #363, opened 2022-11-20
fix: keep existing vehicle config when asleep/offline

carleeno:
Vehicle config is available in the vehicles list api, which is available even when the car is offline. However TeslaCar was referencing it from vehicle data api, which is not always available. By moving this, it ensures that any config values are always available.

Notably:
- Rear seat heaters (including 3rd row seats) availability is available when the vehicle is asleep, thus if HA is restarted while the vehicle is asleep, the rear seat heater entities don't go missing
  - This fixes https://github.com/alandtse/tesla/issues/297
- power lift gate availability is always available
  - this ensures that cover.close for the trunk remains available when the vehicle is asleep, but I don't have a PLG to validate this
- heated steering wheel availability determination was fixed to not confuse "off" with "unavailable"
  - This should fix https://github.com/alandtse/tesla/issues/283, but I don't have a heated steering wheel to validate this

Unfortunately this does NOT fix the following entities becoming unavailable when the integration is reloaded (or HA restarted) while the vehicle is asleep:
  - sentry mode
  - homelink
  - steering wheel heater

A different solution would need to be implemented for those, as they rely on vehicle data api. One idea is to always create the entities, but set their `_attr_entity_registry_enabled_default` to false and let users manually enable them if they have those features and want to use them. After a reload, those entities would remain enabled if the user enabled them previously.


carleeno:
I really should have checked tests before pushing this lol.
Will fix shortly...

carleeno:
@shred86 for vis, this fixes some of the vehicle asleep entity loss, but not all

alandtse:
Wow. Thanks for hunting that down.

teslajsonpy/car.py
```diff
@@ -485,7 +486,7 @@ def software_update(self) -> dict:
     @property
     def steering_wheel_heater(self) -> bool:
         """Return steering wheel heater option."""
-        return self._vehicle_data.get("climate_state", {}).get("steering_wheel_heater")
+        return self._vehicle_data.get("climate_state", {}).get("steering_wheel_heater") is not None
```
alandtse, line 489: Does this mean the steering_wheel may not have an unavailable state? I realize the typing says `bool`, but if we need to convey unavailability the fix is to make it `bool | None`

carleeno: `steering_wheel_heater` is only used for "is it available"; and `is_steering_wheel_heater_on` is used for on/off determination.

I thought about renaming this property to `steering_wheel_heater_available` but that would break backwards compatibility with the tesla integration, in case that matters.

alandtse: Ok that's fine. I didn't realize it was just feature availability.

rlippmann:
Just curious, from the logs I posted in the issue thread, api_version is 42, here it’s 36.  Does that make a difference?

carleeno:
> Just curious, from the logs I posted in the issue thread, api_version is 42, here it’s 36. Does that make a difference?

the version 36 is just from some old mock data, but it does bring up a good point that older api versions might not have the vehicle_config in the vehicles api (only the vehicle state api).

My car is still on 2022.20.* and has api version 42 as well, so unless someone is purposely holding back to a very old software version they shouldn't have any problems.

alandtse:
Perhaps this means we should update the mock data to something newer.

PR #365, opened 2022-11-22
fix: graceful handling of unavailable sites/vehicles.

carleeno:
This should fix a couple of related issues:
- https://github.com/alandtse/tesla/issues/324
- https://github.com/alandtse/tesla/issues/315

If `get_vehicle_data` failed for a TeslaException that wasn't a 408, it would try to return an unassigned variable. Now it will re-raise that "unexpected" exception to let the caller handle it.

Additionally the calls for `get_vehicle_data` in `generate_car_objects` did not handle a vehicle being offline gracefully, and would fail the entire integration setup. The same was for `generate_energysite_objects` for a very similar reason.

Now if the vehicle/site data is unavailable during setup, the device will still be added with the data dict empty. Each of the car/site classes are designed to handle an empty data dictionary, so this should be fine. From what I can tell, it's not really possible to NOT add a device that's in the vehicle list or site list without refactoring a lot of controller.py.

All of these cases are logged as warnings, as they don't break anything, but would be unusual/unexpected.

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -506,8 +506,9 @@ class Controller:
 
         except TeslaException as ex:
             if ex.message == "VEHICLE_UNAVAILABLE":
-                _LOGGER.debug("Vehicle asleep - data unavailable.")
+                _LOGGER.debug("Vehicle offline - data unavailable.")
                 return {}
+            raise ex
 
         return response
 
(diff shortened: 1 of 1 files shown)


alandtse (approved):

alandtse:
Please note that we still need a verb after the `fix:`. I know it's weird, but it's consistent with other stuff like `feat:`

carleeno:
> Please note that we still need a verb after the `fix:`. I know it's weird, but it's consistent with other stuff like `feat:`

ahh thanks, that makes sense

PR #378, opened 2022-12-08
fix: Don't wake car for pointless commands

carleeno:
Closes https://github.com/alandtse/tesla/issues/379

This strategically checks if some commands would be pointless if the car is asleep, and sets `wake_if_asleep` accordingly.

Also refactors default parameters to save ~100 lines

(this is split out from #373)

### DIFF
--- a/teslajsonpy/car.py
+++ b/teslajsonpy/car.py
@@ -650,9 +650,26 @@ class TeslaCar:
         )
 
     async def _send_command(
-        self, name: str, *, path_vars: dict, wake_if_asleep: bool = False, **kwargs
+        self,
+        name: str,
+        *,
+        additional_path_vars: dict = None,
+        wake_if_asleep: bool = True,
+        **kwargs,
     ) -> dict:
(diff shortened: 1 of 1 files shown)


carleeno:
wasn't sure if this was a fix or a feat btw, feel free to change

teslajsonpy/car.py
```diff
@@ -703,53 +719,31 @@ async def change_charge_limit(self, value: float) -> None:
 
     async def charge_port_door_close(self) -> None:
         """Send command to close charge port door."""
-        data = await self._send_command(
-            "CHARGE_PORT_DOOR_CLOSE",
-            path_vars={"vehicle_id": self.id},
-            wake_if_asleep=True,
-        )
+        data = await self._send_command("CHARGE_PORT_DOOR_CLOSE")
```
alandtse, line 722: Why are we no longer specifying to wake it if asleep?

alandtse: Nevermind, you're using the default. Ok.

PR #379, opened 2022-12-09
fix: Improve wake up retry logic

carleeno:
The wake_up decorator was removed as it was only used for the `api()` method. The `wake_if_asleep` logic has been refactored between the `Controller.wake_up` and `Controller.api` methods and made to be more readable and handle edge cases better:
- If api is called for a endpoint that does not accept a vehicle_id, and `wake_if_asleep=True`, it will raise an exception
  - with the decorator, it would have failed as it's not possible to wake a vehicle without it'd ID, but also it's not needed to wake a vehicle for these types of commands
- If the api is called with `wake_if_asleep=True` and the car has gone to sleep so recently that the Controller doesn't know it yet, the initial api request will fail, but now it will handle that gracefully, wake the vehicle, and try the command again
  - it appeared the decorator was supposed to do this, but due to some misplaced conditions, it did not.

It's also more efficient by only retrying the wake command if it failed, otherwise it waits for it to come online by polling the `VEHICLE_SUMMARY` endpoint at regular intervals with it's own timeout.

(split out from #373)

Tested with HA dev container and model 3
- ensured sending a command with `wake_if_asleep=True` (climate on) while car was asleep would wake the car, only sending 1 wake request, and then detect that the car waked (it took ~8 seconds)
- ensured sending a command with `wake_if_asleep=False` (climate off) while car was asleep would not wake the car, and logs would not show an error that the car was unavailable
- ensured that sending a command with `wake_if_asleep=True` while car was awake would not try to wake the car


carleeno:
(in draft while I finish some real world testing)

PR #383, opened 2022-12-10
fix: correct typo in wake_up call

carleeno:
Fix a bug introduced in #379

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -1346,7 +1346,7 @@ class Controller:
             # It may fail if the car slept since the last api update
             if not valid_result(response):
                 # Assumed it failed because it was asleep and we didn't know it
-                await self.wake_up
+                await self.wake_up(car_id=car_id)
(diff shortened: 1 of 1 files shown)


carleeno:
@alandtse sorry about this, I totally missed it. Fortunately it's not common to reach this line of code (the car has to have gone offline since the last `update` when sending a command). But we probably should point release teslajsonpy for this fix before we release 3.5.0 because with enough users out there, we will see this bug pop up.

alandtse:
I'm on mobile so can't create a PR. Can you do a PR from Dev to master please? I just date those and ignore the semantic versioning requirement.

carleeno:
> I'm on mobile so can't create a PR. Can you do a PR from Dev to master please? I just date those and ignore the semantic versioning requirement.

Unfortunately it seems I'm blocked from making a PR against master. however the tesla integration doesn't have this bug yet, so it probably doesn't have to be rushed, I just wanted to make sure you saw this before cutting the next tesla integration release.

PR #390, opened 2022-12-29
fix: handle option codes unavailable

InTheDaylight14:
related to https://github.com/alandtse/tesla/issues/466


alandtse:
Great. Thanks!

PR #392, opened 2023-01-12
fix: Replace charge_amps with charge_current_request in set_charging_amps

Bre77:
When calling `set_charging_amps` it looks for a parameter `charge_amps` and uses this to save the updated value, however this parameter isnt used anywhere else. I believe this is actually meant to be the `charge_current_request` parameter.

Fixes https://github.com/alandtse/tesla/issues/479

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -19,3 +19,4 @@
 -   carleeno [Github](https://github.com/carleeno)
 -   shred [Github](https://github.com/shred86)
 -   InTheDaylight14 [Github](https://github.com/InTheDaylight14)
+-   Bre77 [Github](https://github.com/Bre77)
--- a/teslajsonpy/car.py
+++ b/teslajsonpy/car.py
@@ -897,7 +897,10 @@ class TeslaCar:
             )
 
         if data and data["response"]["result"] is True:
-            params = {"charge_current_request": int(value)}
+            params = {
+                "charge_amps": int(value),
+                "charge_current_request": int(value),
+            }
             self._vehicle_data["charge_state"].update(params)
 
     async def set_cabin_overheat_protection(self, option: str) -> None:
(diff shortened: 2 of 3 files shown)


Bre77:
To add this this, there doesn't appear to be any way to access the `charge_amps` parameter, so it could probably be completely removed, including the testing fixtures.

alandtse:
Thanks. Can you please confirm the API call actually works as expected? You can test with the custom api in HA. Please also consider adding a test for this so future changes don't revert this.

Lastly, mainly optional, we may want to submit a change to the unofficial API if Tesla has indeed changed the parameter. https://tesla-api.timdorr.com/vehicle/commands/charging#post-api-1-vehicles-id-command-set_charging_amps

alandtse:
Test failure is unrelated to your change and can be ignored. But please note the request to update the test.

Bre77:
@alandtse if the correct parameter is `charging_amps` then would the fix instead be adding a public method to access it in the library, and for you to update the HACS integration to use that instead of `charge_current_request`? Right now the value that is set and the value that are read are completely different. (charging_amps/charge_current_request)

alandtse:
Please do verify. There are state getters and commands to set stuff. You've linked the state command but I also provided the last known command. It's totally possible it's changed too.

Bre77:
> Please do verify. There are state getters and commands to set stuff. You've linked the state command but I also provided the last known command. It's totally possible it's changed too.

Setting this value has not changed/broken in the HACS integration, I am using your integration succesfully to change this value already. It's only the _state_ not being updated that is broken.

Bre77:
I have confirmed that the "CHARGING_AMPS" command updates both "charge_amps" and "charge_current_request", so this PR is still valid.

`"charge_state": {
        "battery_heater_on": False,
        "battery_level": 79,
        "battery_range": 277.34,
        "charge_amps": 1,
        "charge_current_request": 1,
        "charge_current_request_max": 32,
        "charge_enable_request": True,`

Bre77:
For potential backwards compatability, and to match how the Tesla API behaves, I have changed `set_charging_amps` to set both **charge_amps** and **charge_current_request** in _vehicle_data

alandtse:
Thanks!

PR #402, opened 2023-03-26
feat: add energy_site_ids and vins to update to limit updating to specific devices

bdraco:
When a user has multiple energy sites/cars they all update even if the device is disabled in Home Assistant

Related PR https://github.com/alandtse/tesla/pull/552


alandtse:
Thanks for the whole set.

PR #406, opened 2023-05-01
feat: Add media controls

Megabytemb:
This adds abstractions and the adjust volume endpoint


teslajsonpy/car.py
```diff
@@ -35,6 +35,18 @@
 }
 
 
+def safeget(dct, *keys, default=None):
+    """Get a recursuive object from a dict."""
+    for key in keys:
+        try:
```
alandtse, line 41: Why not use dict.get?

teslajsonpy/car.py
```diff
@@ -35,6 +35,18 @@
 }
 
 
+def safeget(dct, *keys, default=None):
+    """Get a recursuive object from a dict."""
```
alandtse, line 39: `recursive`

alandtse:
Just to confirm, did you still want to do this?

teslajsonpy/car.py
Megabytemb: fixed

teslajsonpy/car.py
Megabytemb: I don't use get, otherwise we end up with `.get(...).get(...).get(....)` and it can get quite long and messy.

This little helper function cleans that up.
If you're happy with it, in a later pull request, i can update all the others to use the helper function.

alandtse:
Btw, I'm not sure why this touches every file in the repo. This will really muddle history. Can you figure out why?

PR #411, opened 2023-06-14
perf: avoid bytes to text conversion overhead

bdraco:
orjson can decode bytes directly without the need to convert it to text first.

Avoid calling `.text` since it has to do extra processing under the hood in debug as well

### DIFF
--- a/teslajsonpy/connection.py
+++ b/teslajsonpy/connection.py
@@ -188,19 +188,21 @@ class Connection:
         if not baseurl:
             baseurl = self.baseurl
         url: URL = URL(baseurl).join(URL(url))
-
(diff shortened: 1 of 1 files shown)


alandtse:
Just to confirm, you've tested this to make sure our API use isn't expecting text responses? I can't remember how we designed this.

bdraco:
I've been running this in production on two systems and all good.

`orjson.loads` takes any of `bytes` `bytearray` `memoryview` `str` https://github.com/ijl/orjson#deserialize so it should handle text or bytes just fine

alandtse:
Thanks for confirming. Perf won't trigger a release but apparently we have a bug that needs to be squashed so I think we'll have a release soon.

PR #431, opened 2023-10-11
fix: handle energysites without powerwall

gkwok1:
https://github.com/alandtse/tesla/issues/738


alandtse:
Thanks. Please note that to take this in tesla, I'm going to have to ask you try to fix all the tests that broke with the change to the API. I didn't force it before since people were down, but right now I've stopped accepting feature pull requests until we get back to a sane state with the tests.

Issue #448, opened 2024-01-23
Change to https://owner-api.teslamotors.com/api/1/vehicles API


grillp:
teslajsonpy Version: 3.9.9

Seems Tesla has changed access to the https://owner-api.teslamotors.com/api/1/vehicles call.. This means that calls to `vehicle_list()` now fails with:

`requests.exceptions.HTTPError: 412 Client Error: Endpoint is only available on fleetapi. Visit https://developer.tesla.com/docs for more info for url: https://owner-api.teslamotors.com/api/1/vehicles`

It appears you have to use `https://owner-api.teslamotors.com/api/1/products` now and parse out all vehicles..

sample app:
```python
import teslapy
token = "<tesla api token>"
with teslapy.Tesla('elon@tesla.com') as tesla:
    if not tesla.authorized:
        tesla.refresh_token(refresh_token=token)
    vehicles = tesla.vehicle_list()
    print(vehicles[0])
```
results:
```
Traceback (most recent call last):
  File "/Users/me/work/personal/teslapy/test_tesla.py", line 6, in <module>
    vehicles = tesla.vehicle_list()
               ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/teslapy/__init__.py", line 375, in vehicle_list
    return [Vehicle(v, self) for v in self.api('VEHICLE_LIST')['response']]
                                      ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/teslapy/__init__.py", line 369, in api
    return self.request(endpoint['TYPE'], uri, serialize,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/teslapy/__init__.py", line 158, in request
    response.raise_for_status()  # Raise HTTPError, if one occurred
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 412 Client Error: Endpoint is only available on fleetapi. Visit https://developer.tesla.com/docs for more info for url: https://owner-api.teslamotors.com/api/1/vehicles
```

jaymunro:
[...]

PR #450, opened 2024-02-02
fix: Remove references to deprecated vehicles endpoint.

iancg:
This should resolve issue #448 for pre-2021 Model X/Model S cars.

Based on the changes described in https://github.com/alandtse/tesla/issues/774#issuecomment-1917821933

It won't necessarily help with newer cars, but should do no harm.

Tested with 2017 MX.


alandtse:
Thanks!

PR #454, opened 2024-02-20
feat: Custom url no sig changes

thierryvt:
defaults to using the vehicle VIN by checking (and overwriting) the vehicle_id path param and without changing any of the function signatures


thierryvt:
@alandtse @llamafilm have a look when you've got a minute. This approach avoids changes to any of the functions or their signatures as it lays the responsibility to ensure the VIN is used inside the function that handles the api communication itself.

I've set it to draft as I still need to verify this actually works, all the tests pass but consider it a proof of concept until then.

alandtse:
Please address the lint issues.

thierryvt:
done, still trying to verify it's working though. My pc isn't co-operating today...

thierryvt:
@alandtse needed a minor fix but verified this works for my car, so as long as the Tesla guy wasn't lying about the old API supporting VINs as IDs this should work for everyone with 0 impact.

llamafilm:
I like this approach, preserving compatibility.  I tested this and it works with my car.  I suggest a few changes:

- squash all these commits into one
- Bump version to 3.10 in CHANGELOG.md, pyproject.toml, and __version__.py
- Remove the "Tesla has no official API" warning in README.md

One thing to note: Reinstalling a new version of the add-on (https://github.com/thierryvt/tesla) which depends on this updated library did not actually update the library.  I had to login to the homeassistant Docker container and manually reinstall the newer teslajsonpy.  Hopefully that's just something weird about my dev environment and it will not affect other users...

alandtse:
Do not bump versions. You do not control that. It is controlled by conventional commits.

thierryvt:
I'll squash the commits and remove the warnings.

I had the same experience with trying to test this code btw, it should work fine when this gets released and the version bumped in the tesla plugin.

thierryvt:
bit messy because pycharm has a different idea of what squashing means than I do. Nothing a bit of force pushing and head resetting can't solve.

Unless there's more feedback this PR is ready to merge.

alandtse:
We squash on merge so you didn't need to squash anything. But ok.

PR #456, opened 2024-02-23
fix: maximize chance of reusing the connection while updating

bdraco:
Previously we would gather which would open a flood of connections and since the cost of establishing the TLS connection `do_handshake` was higher than the actual request, it would take more wall clock time to update and was more resource intensive.


alandtse:
Thanks. Did you notice any 408s or connection errors previously? I may consider it a fix in that case.

bdraco:
It went unavailable frequently before this change, but I didn't realize that until I looked at the history that they went away so I didn't put it in the comments as I didn't connect it until you asked

alandtse:
Thanks. That's good enough for me to make it a fix.

Issue #462, opened 2024-03-27
TeslaCar object does not have "vehicle_config"


kbrint:
# Background

In https://github.com/alandtse/tesla/issues/860 we noticed that a recent change made it impossible to close the trunk via the API.

Follow-up discussion on the pull request https://github.com/alandtse/tesla/pull/868 shows that the problem is likely in teslajsonpy.

# Cause

While debugging, we find that the check for `.get("vehicle_config",{}).get("plg")` in `TeslaCar.powered_lift_gate` never returns True.  It's because `self._car` never contains a key for `vehicle_config`.

I think the problem is that the `TeslaCar._car` only contains the data from the `PRODUCT_LIST` query.   The data stored in my TeslaCar object has the following keys: `dict_keys(['id', 'user_id', 'vehicle_id', 'vin', 'color', 'access_type', 'display_name', 'option_codes', 'cached_data', 'granular_access', 'tokens', 'state', 'in_service', 'id_s', 'calendar_enabled', 'api_version', 'backseat_token', 'backseat_token_updated_at', 'ble_autopair_enrolled', 'command_signing'])`

# Fix

I think the fix for `powered_lift_gate` is to look for the `vehicle_config` in the `self.._vehicle_data` dict.

kbrint:
#463 should fix this.

PR #463, opened 2024-03-27
fix: Use the correct location for `vehicle_config`.  (#462)

kbrint:


kbrint:
I tried patching in the test only (to the previous dev branch) and they still all pass.

I think that is a separate bug, and that's possibly how this error slipped into the code in the first place.

alandtse:
Thanks. I mean it's weird cause it reverts the prior change but we can see if it results in a revert war.

kbrint:
> Thanks. I mean it's weird cause it reverts the prior change but we can see if it results in a revert war.

The change in question (92774e9) modifies references to:
 *  `vehicle_config` (the subject of this PR), and
 * `vehicle_state`, and
 * `climate_state`.

It appears that the `vehicle_config` contains immutable attributes of the vehicle (as opposed to the `*_state` entries).  I don't think it needs to be refreshed often, nor should the values change if the car is offline.  Anyway, this is where the data actually is.

PR #466, opened 2024-04-04
fix: fix grid_status

ericdegroot:
I recently had a Powerwall 3 installed, and I have no local API access. I'm using [Tesla Custom Integration](https://github.com/alandtse/tesla) to integrate the battery into my Home Assistant setup via the owner cloud. The grid status binary sensor provided by that integration always inaccurately showed as 'Clear' or `off`. I traced this back to the `PowerwallSite.grid_status` property in this project. Changing this property to lookup the value in `self._site_data` instead of `self._site_summary` fixed the issue for me. Your test mock data suggests that site data was always the correct source, but I do not have access to any other powerwalls to see if this is unique to Powerwall 3 cloud access.


alandtse:
Thanks!

PR #470, opened 2024-06-03
feat: add cooling seats

dandelionclock:
[Issue 977](https://github.com/alandtse/tesla/issues/977) under Tesla Custom Integration

Added ability to retrieve states of and change state of cooled seats


alandtse:
Please don't mix and match PRs. Use separate branches for different things.

dandelionclock:
done thanks

PR #477, opened 2024-11-06
fix: json attribute error for GET with parameters

ehendrix23:
When doing an API call for which the method is get but where parameters are provided an exception occurs for unknown attribute json=data.
This is due to the  httpx get method not supporting the json parameter, instead the params parameter has to be provided.

This fixed checks if the method is get and if it is will provide the additional parameters through params parameter instead.

### DIFF
--- a/teslajsonpy/connection.py
+++ b/teslajsonpy/connection.py
@@ -200,9 +200,14 @@ class Connection:
 
         try:
             if data:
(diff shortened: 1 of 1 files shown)


ehendrix23:
Fixed the title issue.

Not sure why 3.10 failed the test as I do not see how my small change would cause it to fail.

Issue #481, opened 2025-02-04
Request Configurable driving polling interval for Cost Optimization


flaviorighi:
I wanted to bring up a concern related to API usage costs. After recent changes in Tesla's pricing model, optimizing API calls has become more critical than before to manage costs effectively.

This module currently uses a hardcoded `DRIVING_INTERVAL` value, which determines how frequently API requests are made while driving. This fixed interval prevents users of the module, like [Tesla Home Assistant](https://github.com/alandtse/tesla) integration (and potentially other applications) from optimizing API costs based on their specific needs.

Since this parameter is critical for balancing API usage and real-time functionality, it should be configurable via user settings. This would allow users to adjust the interval as needed (e.g., reducing frequency for cost-saving or increasing it for more frequent updates).

It would be great to add another controller constructor parameter like `_driving_interval` to let this externally configurable.

PR #483, opened 2025-02-22
feat: added configurable parameter driving_interval

flaviorighi:
Related to issue: #481 

Extras:
- Removed useless DRIVER_INTERVAL usage in connection.py

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -24,3 +24,4 @@
 -   thierryVT [Github](https://github.com/thierryvt)
 -   llamafilm [Github](https://github.com/llamafilm)
 -   ericdegroot [Github](https://github.com/ericdegroot)
+-   flaviorighi [Github](https://github.com/flaviorighi)
--- a/poetry.lock
+++ b/poetry.lock
@@ -222,26 +222,26 @@ files = [
 
 [[package]]
 name = "attrs"
-version = "23.2.0"
+version = "24.2.0"
 description = "Classes Without Boilerplate"
 optional = false
 python-versions = ">=3.7"
 groups = ["main"]
 files = [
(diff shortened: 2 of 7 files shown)


alandtse:
Can you provide more context on the change? I'm not certain the parameter was useless. We're you saying it was broken so this is a fix?

flaviorighi:
The original code had a slightly different structure, I simplified the condition for better readability. The updated version:
```python
while not (
    disconnected
    or time.time() > last_message_time + WEBSOCKET_TIMEOUT  # 11s
    or time.time() > last_message_time + DRIVING_INTERVAL  # 60s
):
```
makes it clearer by directly expressing the timeout conditions in a more readable way. However, the last condition time.time() > last_message_time + DRIVING_INTERVAL is unnecessary because if time.time() > last_message_time + WEBSOCKET_TIMEOUT (11s) is met, then the DRIVING_INTERVAL condition (60s) will always be met later. In other words, the second condition already ensures termination before the last one becomes relevant, making it redundant.

That said, if the values of WEBSOCKET_TIMEOUT and DRIVING_INTERVAL were to change in the future, we could consider keeping only the condition with the minimum value to ensure the earliest termination.

flaviorighi:
I've removed the python version 3,8 from `test_and_lint` action and bumped to 3.10 in `semantic_relase` because [ poetry dropped 3,8 support](https://github.com/python-poetry/poetry/pull/9692) since v2.0.0.

alandtse:
Please don't overload the PR. The commit message should basically describe the change.  Let's split the configurable driving interval from the underlying Python bump.

flaviorighi:
I had to update verions, all the actions were failing due that.
Do you prefer to do another pr in the middle just to have things working?

alandtse:
Can you do a PR with the python changes first and we merge that and you rebase?

flaviorighi:
> Can you do a PR with the python changes first and we merge that and you rebase?

Opened #484

PR #487, opened 2025-07-21
fix: Improve testing of result and adjust for Tesla API change

ChrisJ60:
Existing test can return incorrect result (fail instead of pass) when HTTP status code is e.g. 200 or some other '2xx' value (all indicate 'okay'). Tesla API has been changed so that (a) result code is now almost always '200' and the 'code' element in the response has changed to 'Code'. Don't ask me why.

### DIFF
--- a/teslajsonpy/energy.py
+++ b/teslajsonpy/energy.py
@@ -203,7 +203,7 @@ class PowerwallSite(EnergySite):
             path_vars={"site_id": self.energysite_id},
             default_real_mode=real_mode,
         )
-        if data and data["response"]["code"] == 201:
+        if data and data["response"] and data["response"]["Code"] and (len(str(data["response"]["Code"])) == 3) and (str(data["response"]["Code"])[:1] == "2"):
             self._site_config.update({"operation": real_mode})
 
     async def set_reserve_percent(self, value: int) -> None:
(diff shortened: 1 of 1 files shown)


teslajsonpy/energy.py
```diff
@@ -203,7 +203,7 @@ async def set_operation_mode(self, real_mode: str) -> None:
             path_vars={"site_id": self.energysite_id},
             default_real_mode=real_mode,
         )
-        if data and data["response"]["code"] == 201:
+        if data and data["response"] and data["response"]["Code"] and (len(str(data["response"]["Code"])) == 3) and (str(data["response"]["Code"])[:1] == "2"):
```
alandtse, line 206: Thanks. This is quite complex and brittle if the keys change again.

1. First use dict.get(). data.get("response") will be None if the key doesn't exist. You can nest it further. 
```python
code = data.get("response", {}).get("Code", 0) or data.get("response", {}).get("code", 0) 
```
2. If you're dealing with http codes, then just use number ranges that are valid. e.g., 
```python
code >=200 and code < 300
```
, or whatever you're looking for.

Double check the syntax, I've been coding cpp all day so may have forgotten.

alandtse (changes requested):
I realize you're fixing someone else's broken code, but part of the reason it's brittle is the direct access to keys that may or may not exist. See my suggestions below. I think you can run it through ChatGpt to fix the syntax and even explain what I just provided.

alandtse:
#488 should handle

PR #488, opened 2025-07-25
fix: Fix bug Tesla caused by mixing the case of the field "code" in their responses

djbadders:
Fix bug Tesla caused by mixing the case of the field "code" in their responses, where it can be "code" or "Code" depending on the call. This has caused the energy integration to break when setting values such as the reserve and charging from grid.


teslajsonpy/energy.py
```diff
@@ -203,7 +203,7 @@ async def set_operation_mode(self, real_mode: str) -> None:
             path_vars={"site_id": self.energysite_id},
             default_real_mode=real_mode,
         )
-        if data and data["response"]["code"] == 201:
+        if data and (data["response"].get("code") == 201 or data["response"].get("Code") == 201):
```
alandtse, line 206: Let's also add a get check on response.

alandtse (changes requested):
Thanks.

djbadders:
Yes I have a powerwall and without this change it's updates but we get an error

alandtse:
Ok in tesla we'll need to bump to get this version.

PR #502, opened 2026-07-27
fix: don't assume the SITE_DATA response is a dict

LuShadowX:
`get_site_data()` returns `["response"]` straight from the API, and `connection.py` only raises `TeslaException` for `status_code > 299` or a body with an `error` key. An energy site that answers HTTP 200 with `"response"` as a bare string therefore flows through untouched, and `_site_data[energysite_id]` becomes a `str`.

Downstream that shows up as:

- `SolarSite.data_available` is `self._site_data != {}`, and a non-empty string passes, so the site looks like it has data;
- `solar_power` / `grid_power` / `load_power` then hit `'str' object has no attribute 'get'`;
- on the next poll `_get_and_process_site_data` hits `'str' object has no attribute 'update'` — alandtse/tesla#494, open since 2023.

Recent report with a full traceback: https://github.com/alandtse/tesla/discussions/1231 (a leased SolarCity site; the owned site on the same account is fine).

This returns `{}` and logs a warning when the payload isn't a dict, so `data_available` correctly reports `False` and the site is skipped instead of half-created.

Scoped deliberately to `get_site_data`, since that is the path with a confirmed report. `get_site_config` and `get_site_summary` have the same shape and could get the same guard if you'd like it here.

Added a test that reproduces it — it fails on `dev` with `assert 'site_unavailable' == {}` and passes with the fix. Full suite: 74 passed.

### DIFF
--- a/teslajsonpy/controller.py
+++ b/teslajsonpy/controller.py
@@ -359,9 +359,20 @@ class Controller:
 
     async def get_site_data(self, energysite_id: int) -> dict:
         """Get site data json from TeslaAPI for a given energysite_id."""
-        return (await self.api("SITE_DATA", path_vars={"site_id": energysite_id}))[
+        response = (await self.api("SITE_DATA", path_vars={"site_id": energysite_id}))[
             "response"
         ]
+        if not isinstance(response, dict):
+            # Some energy sites answer HTTP 200 with a bare string instead of an
+            # object. That is not a TeslaException, so it would otherwise be stored
+            # as-is and later crash on .get()/.update().
+            _LOGGER.warning(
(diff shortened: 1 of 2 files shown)


alandtse (approved):

LuShadowX:
The `build (3.9)` failure isn't related to this change - it fails at the `Install Poetry` step, before Black/Lint/pytest run. `snok/install-poetry` installs the latest Poetry, and Poetry dropped 3.9 in 2.3.0 (`requires_python >=3.10`), so the 3.9 job can't install it. Same job was already failing before this PR.

Happy to send a separate PR pinning Poetry if useful.

LuShadowX:
Thanks for the quick review and merge.Happy to help!😁

PR #503, opened 2026-07-27
ci: pin Poetry to 2.2.1

LuShadowX:
`build (3.9)` has been failing at the `Install Poetry` step, so Black, Lint and pytest never run on 3.9.

`snok/install-poetry` installs the latest Poetry, and Poetry raised `requires_python` to `>=3.10` in 2.3.0. Pinning to 2.2.1, the last release that still supports 3.9.

If you'd rather drop 3.9 from the matrix instead, happy to switch.

### DIFF
--- a/.github/workflows/test_and_lint.yaml
+++ b/.github/workflows/test_and_lint.yaml
@@ -31,6 +31,7 @@ jobs:
     - name: Install Poetry
       uses: snok/install-poetry@v1
       with:
+        version: "2.2.1"
         virtualenvs-create: true
         virtualenvs-in-project: true
     #----------------------------------------------


LuShadowX:
@alandtse this is the Poetry pin I offered in #502 — the `Lint and test` run here is waiting on your approval (fork PR touching `.github/workflows`), so it can't prove itself until then. Once it runs, the 3.9 job should get past `Install Poetry` again; `2.2.1` is the last release that still allows Python 3.9.

alandtse:
We should deprecate 3.9 and not pin on the ci.

LuShadowX:
Agreed, and the pin was only ever a stopgap. Happy to help!😁

